Issue #19, opened 2025-03-12
How to support cursor


tuskermanshu:
After using the demo, it works fine, but when I put it into the cursor's mcp, it gives me an error.

````
 Error connecting to server: McpError: MCP error -32601: Method not found
    at Server._onresponse (file:///Users/tuskermanshu/demo/cmp- demo/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:203:27 )
    at _transport.onmessage (file:///Users/tuskermanshu/demo/cmp-demo/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:75:22)
    at SSEServerTransport.handleMessage (file:///Users/tuskermanshu/demo/cmp-demo/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/dist/esm/server/sse.js:93:71)
    at SSEServerTransport.handlePostMessage (file:///Users/tuskermanshu/demo/cmp-demo/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/dist/esm/server/sse.js:72:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Server.<anonymous> (file:///Users/tuskermanshu/demo/cmp-demo/node_modules/.pnpm/mcp-proxy@2.10.5/node_modules/mcp-proxy/dist/chunk-RZVGSMPC.js:148:7) {
  code: -32601,
  data: undefined
}
node:_http_server:351
    throw new ERR_HTTP_HEADERS_SENT('write');
          ^

Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client
    at ServerResponse.writeHead (node:_http_server:351:11)
    at Server.<anonymous> (file:///Users/tuskermanshu/demo/cmp-demo/node_modules/.pnpm/mcp-proxy@2.10.5/node_modules/mcp-proxy/dist/chunk-RZVGSMPC.js:129:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  code: 'ERR_HTTP_HEADERS_SENT'
}

Node.js v23.7.0
````

PR #21, opened 2025-03-14
fix: Cursor errors when listing roots

TylerBarnes:
Hey! I'm working on a docs MCP for Mastra.ai (https://github.com/mastra-ai/mastra/pull/2966) and ran into the issue that it doesn't work in Cursor.

Seems the issue is we need to check if the client supports listing roots changes, not just if it supports roots.
I'm not 100% sure about the change so interested to hear your thoughts. If I do a try/catch here without checking for `roots.listChanges` that also works. Any idea the difference between the client supporting roots vs the client supporting roots.listChanges?

Fixes https://github.com/punkpeye/fastmcp/issues/19

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -568,10 +568,13 @@ export class FastMCPSession<T extends FastMCPSessionAuth = FastMCPSessionAuth> e
       console.warn('[warning] FastMCP could not infer client capabilities')
     }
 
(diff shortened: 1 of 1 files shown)


TylerBarnes:
Published my fork as `tylerbarnes-fastmcp-fix` if anyone needs to get unblocked in cursor

punkpeye:
Thank you

PR #22, opened 2025-03-23
add "extraArgs" for each tool

j4ys0n:


punkpeye:
@j4ys0n Anything I can help with here?

j4ys0n:
@punkpeye I clicked too fast - didn't mean to open a PR against your repo. I wanted to fully try it out first. That said, I did, and it works as expected. If you're cool with adding this feature I'll open a new PR.

I left the validated args as-is and made it so that any args that are passed that aren't in the specified schema, get moved to `extraArgs` and are available to the tool's `execute` handler.

I added this because I want to be able to pass sensitive arguments, like tokens, api keys, etc but I don't want them to go through the LLM, thus, they're not in the tool's schema that gets exposed to the LLM.

punkpeye:
Let's do it. It looks valuable

PR #23, opened 2025-03-24
put "unvalidated" args in separate argument during argument validation

j4ys0n:
When tool arguments are validated against the defined schema, instead of removing arguments that aren't in the schema, this will move them to a separate argument called `extraArgs` this is available in the tools' execution handler. This was done to be able to handle sensitive values like api keys or authentication tokens without adding them to the schema that the LLM will use.


punkpeye:
@j4ys0n as I am reviewing this and thinking through it, would it not make more sense to pass those hidden values in headers?

punkpeye:
Alternatively, if the goal is to get extra params without defining them in the schema, could you not just change the schema to use [`.passthrough()`](https://zod.dev/?id=passthrough) and get the extra params that way?

j4ys0n:
Apologies for the delay. Yeah, could do `passthrough`. Up to you. I kind of like having the args separate for some reason but I don't have a great reason for it.

punkpeye:
Sorry, I should have caught this earlier, but I think `.passthrough()` is the appropriate API to use. Unnamed parameters magically appearing as extra params will create confusion for new users.

PR #27, opened 2025-04-01
Switch from Zod to Standard Schema

### LINKED ISSUE #26: Replace Zod with Standard Schema for validation
## Background

FastMCP currently uses Zod directly for schema validation. However, the Standard Schema specification has been developed as a common interface for JavaScript/TypeScript schema libraries. The creators of Zod (along with Valibot and ArkType) designed this specification to make it easier for ecosystem tools to accept user-defined type validators without needing custom adapters for each library.

## Request

Please consider replacing direct Zod usage with Standard Schema in FastMCP. This would allow users to:

1. Choose their preferred validation library (Zod, Valibot, ArkType, etc.)
2. Benefit from the "integrate once, validate anywhere" philosophy of Standard Schema
3. Follow best practices recommended by schema library authors

## Rationale

From the Standard Schema documentation:
> "The spec was designed by the creators of Zod, Valibot, and ArkType. Recent versions of these libraries already implement the spec."

Many ecosystem tools are already adopting Standard Schema, including tRPC, TanStack Form, TanStack Router, and others. This change would align FastMCP with this emerging ecosystem standard.

## Implementation Notes

The change should be fairly straightforward since:
1. Zod already implements the Standard Schema interface (in v3.24.0+)
2. The Standard Schema interface is designed to be minimal and easy to integrate

## Example

Current implementation:
```ts
server.addTool({
  name: "add",
  parameters: z.object({
    a: z.number(),
    b: z.number(),
  }),
  // ...
});
```

Potential Standard Schema implementation:
```ts
server.addTool({
  name: "add",
  parameters: anyStandardSchema, // Could be Zod, Valibot, ArkType, etc.
  // ...
});
```

Thank you for considering this request!

jacksteamdev:
## Overview
This PR replaces direct Zod usage with Standard Schema support, allowing users to choose their preferred validation library while maintaining existing functionality.

## Changes
- Added `@standard-schema/spec` and `xsschema` dependencies
- Updated `ToolParameters` type to use `StandardSchemaV1`
- Replaced `zodToJsonSchema` with `toJsonSchema`

## Benefits
- Users can now use any schema library that implements Standard Schema (Zod, Valibot, ArkType)
- Aligns with ecosystem direction supported by schema library creators
- Follows the example of libraries like tRPC, TanStack Form, and others adopting Standard Schema

## Todo
- [x] Implement Standard Schema support
- [x] Update README with examples
- [x] Add usage examples for different schema libraries

## Related Issues
Resolves #26

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,8 @@
     "@types/node": "^22.13.5",
     "@types/uri-templates": "^0.1.34",
     "@types/yargs": "^17.0.33",
+    "@valibot/to-json-schema": "^1.0.0",
+    "arktype": "^2.1.16",
     "eslint": "^9.21.0",
     "eslint-plugin-perfectionist": "^4.9.0",
     "eventsource-client": "^1.1.3",
@@ -66,6 +68,7 @@
     "semantic-release": "^24.2.3",
     "tsup": "^8.4.0",
     "typescript": "^5.7.3",
+    "valibot": "^1.0.0",
     "vitest": "^3.0.7"
   },
   "tsup": {
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "types": "dist/FastMCP.d.ts",
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.6.0",
+    "@standard-schema/spec": "^1.0.0",
     "execa": "^9.5.2",
     "file-type": "^20.3.0",
     "fuse.js": "^7.1.0",
@@ -29,9 +30,10 @@
     "strict-event-emitter-types": "^2.0.0",
     "undici": "^7.4.0",
     "uri-templates": "^0.2.0",
+    "xsschema": "0.2.0-beta.2",
     "yargs": "^17.7.2",
     "zod": "^3.24.2",
-    "zod-to-json-schema": "^3.24.3"
+    "zod-to-json-schema": "^3.24.5"
   },
   "repository": {
     "url": "https://github.com/punkpeye/fastmcp"
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ npm install fastmcp
 
 ```ts
 import { FastMCP } from "fastmcp";
-import { z } from "zod";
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -22,16 +22,18 @@
...
     "fuse.js": "^7.1.0",
     "mcp-proxy": "^2.10.4",
     "strict-event-emitter-types": "^2.0.0",
     "undici": "^7.4.0",
     "uri-templates": "^0.2.0",
+    "xsschema": "0.2.0-beta.2",
     "yargs": "^17.7.2",
     "zod": "^3.24.2",
-    "zod-to-json-schema": "^3.24.3"
+    "zod-to-json-schema": "^3.24.5"
```
jacksteamdev, line 36: This library is a peer dependency of xsschema and isn't strictly required anymore but I left it because removing it would be a breaking change for Zod users.

jacksteamdev:
@punkpeye Happy to help. Thanks for all your work in the MCP scene!

package.json
qcho: I would remove it. The whole idea behind all of this is not to add a new dependency all-together.
And it won't break in runtime but at the install moment. Seems safe enough.

package.json
```diff
@@ -55,6 +57,8 @@
     "@types/node": "^22.13.5",
     "@types/uri-templates": "^0.1.34",
     "@types/yargs": "^17.0.33",
+    "@valibot/to-json-schema": "^1.0.0",
```
qcho, line 60: adding all of these dependencies seems wrong.
They should be [optional peer dependencies](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta).

jacksteamdev: These are dev dependencies for `src/examples/addition.ts`

ryoppippi: I agree @qcho we need to add @valibot/to-json-schema to optional peer deps.
And, I'd love to add zod/zod-to-json-schema to peer deps

PR #39, opened 2025-04-22
Add the MCP Optional instructions option

JarekToro:
Add the MCP Optional instructions option describing how to use the server and its features.

From the underlying server options of the `@modelcontextprovider/sdk`
```ts
export type ServerOptions = ProtocolOptions & {
    /**
     * Capabilities to advertise as being supported by this server.
     */
    capabilities?: ServerCapabilities;
    /**
     * Optional instructions describing how to use the server and its features.
     */
    instructions?: string;
};
```


punkpeye:
This is great. Could you also contribute a README.md update that reflects the new option?

bierdok:
Is there a reason to allow optional instructions only in the stdio transport type?

Issue #24, opened 2025-03-25
How to debug input validation errors?


bguiz:
I'm using the MCP inspector to invoke multiple MCP methods, which are generally working.

However, one particular MCP method consistency outputs this error:

```text
Error MCP error -32602: MCP error -32602: Invalid someMcpMethodName parameters
```

It is a "handled error" ... so:
- I can only see the above error in the inspector.
- I don't see anything error or stack trace in the terminal output.
- I can't make use of console.log inside of the 

How can I debug in this particular scenario?

PR #44, opened 2025-04-25
fix: (#24) add params validation messages to error message, good for human and llm to better understand the situation

zfflxx:
…human and llm to better understand the situaction


punkpeye:
I am not sure if this is a good practice. Need to research more.

zfflxx:
For example, I have 10 params, claude 3.7 make one mistake in one of the params, but claude don't know which one is incorrect because the too broad error messages, so claude have to try again and again to figure out which one is wrong, but if claude can see the precise error message, then claude can fix the issue in one shot, so it's a must, I think.

punkpeye:
Could you please resolve the conflicts and make the tests pass?

punkpeye:
Thank you!

PR #47, opened 2025-04-30
fix [#37] :  roots capability check to prevent listing roots when server doesn't support it

### LINKED ISSUE #37: List roots not supported" Error with FastMCP 1.21.0 in Programmatic Clients
Hi Team,

I'm experiencing a non-fatal error when using the latest version of FastMCP (1.21.0) with programmatic clients:

```
McpError: MCP error -32600: List roots not supported
```

### Reproduction

This happens with Bright Data's MCP server (https://www.npmjs.com/package/@brightdata/mcp) when used with Google's ADK, LangChain, SmolaAgents, etc.

Interestingly, this error does **not** occur:
- When using Claude Desktop as a client
- When using servers like https://github.com/domdomegg/computer-use-mcp (that utilizes older versions of FastMCP - 1.16)

### What I've Tried

I attempted to explicitly disable the roots capability, but still get the error:

```javascript
let server = new FastMCP({
    name: 'Bright Data',
    version: package_json.version,
    capabilities: {
        roots: null  // Explicitly disable roots capability
    }
});
```

You can easily reproduce this issue with my [[MCP_ADK repo](https://github.com/MeirKaD/MCP_ADK)](https://github.com/MeirKaD/MCP_ADK).

### Questions

1. Is anyone else experiencing this issue with FastMCP 1.21.0?
2. Is there a proper way to handle/disable roots capability in the newer versions?
3. Could this be a regression in FastMCP 1.21.0?

The error is non-fatal (servers still work), but it causes warning messages in client logs and could affect integration with frameworks that handle errors more strictly.

Any help would be appreciated!

MeirKaD:
Hi @punkpeye ,

This change fixes #37 

With this change, the server will only try to list roots if:

1. The client supports the roots capability with list changed  notifications
2. The server itself has declared support for the roots capability

This is more robust and will prevent the error I am seeing when the server explicitly disables roots with roots: null or doesn't declare roots support at all.

Tested it on different Agent frameworks and it worked with no issues.

Let me know what you think.

### DIFF
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -859,6 +859,47 @@ test("session knows about roots", async () => {
   });
 });
 
+test("doesn't attempt to list roots when server explicitly disables roots capability", async () => {
+  await runWithTestServer({
+    server: {
+      capabilities: {
+        roots: null  
+      }
+    },
+    client: async () => {
+      const client = new Client(
+        {
+          name: "example-client",
+          version: "1.0.0",
+        },
+        {
+          capabilities: {
+            roots: {
+              listChanged: true,
+            },
+          },
+        }
+      );
+      
+      client.setRequestHandler(ListRootsRequestSchema, () => {
+        return {
+          roots: [
+            {
+              name: "Frontend Repository",
+              uri: "file:///home/user/projects/frontend",
+            },
+          ],
+        };
+      });
+      
+      return client;
+    },
+    run: async ({ session }) => {
+      expect(session.roots).toEqual([]);
+    },
+  });
+});
+
 test("session listens to roots changes", async () => {
   const clientRoots: Root[] = [
     {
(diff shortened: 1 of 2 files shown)


punkpeye:
@MeirKaD any test that we could add here?

MeirKaD:
I've updated the fix to only skip roots listing when it's explicitly disabled (roots: null), not when it's unspecified. This change should fix the original error without breaking the existing tests.

I also added a new test that ensures we don't attempt to list roots when the server explicitly disables the roots capability. This will prevent this issue from recurring in the future.

MeirKaD:
Still not there unfortunately, leaving the failed test aside for a moment, the fix with `this.#capabilities.roots !== null` still doesn't work for Google ADK as a client, for example.

I will spend more time on this and will update once everything is checked properly.

For now, I will close this PR Thank you for your help though!

PR #57, opened 2025-05-06
feat: configurable ping behavior

### LINKED ISSUE #50: Why does FastMCP ping the server every second?
I tried moving a server that was built on the bare `@modelcontextprotocol/sdk` to FastMCP.  When running the MCP Inspector, I noticed that a `Received message for sessionId <id>` line started showing up every second in the console.  At first I thought I'd done something wrong, or that it was always doing this and I just hadn't noticed.  But it stopped when I went back to the old version.

I see now where the ping is set up in FastMCP, after connecting to the server: https://github.com/punkpeye/fastmcp/blob/main/src/FastMCP.ts#L661

But I don't know why it's necessary, especially for a stdio connection.  Should the ping depend on the transport type?  If there's a benefit with SSE mode, I wonder if the ping should be debounced by tool calls, since if the LLM is making lots of tool calls, I'd assume there's no reason to also ping the server.

fauziralpiandi:
Add ping configuration to `ServerOptions` to reduce excessive logging and improve connection handling. Enable pings by default for SSE/HTTP Streaming while disabling for `stdio`. Allow customization of ping interval and log level

Close #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ A TypeScript framework for building [MCP](https://glama.ai/mcp) servers capable 
 - [Typed server events](#typed-server-events)
 - [Prompt argument auto-completion](#prompt-argument-auto-completion)
 - [Sampling](#requestsampling)
-- Automated SSE pings
+- [Configurable ping behavior](#configurable-ping-behavior)
 - Roots
 - CLI for [testing](#test-with-mcp-cli) and [debugging](#inspect-with-mcp-inspector)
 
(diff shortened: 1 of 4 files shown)


punkpeye:
I appreciate all of your contributions. Very high quality!

fauziralpiandi:
> I appreciate all of your contributions. Very high quality!

✨

fwextensions:
Nice, now I can get rid of `patch-package`!

Issue #58, opened 2025-05-07
Support for mcp-remote


carmilso:
I have tried [mcp-remote](https://github.com/geelen/mcp-remote) with an own MCP server that uses `StreamableHTTPServerTransport` and it works and it's useful because not all MCP clients support http at the moment.

But I have tried it with this simple FastMCP server:
```typescript
import { AuthError, FastMCP } from "fastmcp";
import { z } from "zod";

const server = new FastMCP({
  name: "My Server",
  version: "1.0.0",
  authenticate: (request) => {
    const apiKey = request.headers["x-api-key"];

    if (apiKey !== "123") {
      throw new Response(null, {
        status: 401,
        statusText: "Unauthorized",
      });
    }

    // Whatever you return here will be accessible in the `context.session` object.
    return {
      id: 1,
    };
  },
});

server.addTool({
  name: "sayHello",
  execute: async (args, { session }) => {
    return `Hello, ${session.id}!`;
  },
});

server.start({
  transportType: "httpStream",
  httpStream: {
    endpoint: "/mcp",
    port: 8080,
  },
});
```

Using cursor to run it through mcp-remote:
```json
    "pipedrive-http": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote@latest",
        "--header",
        "x-api-key:123",
        "http://localhost:8080/mcp",
        "--transport",
        "http-only"
      ]
    }
```

But in the console I see this error:
```
2025-05-07 14:54:20.766 [error] user-pipedrive-http: [607337] [Local→Remote] tools/list

2025-05-07 14:54:20.772 [error] user-pipedrive-http: [607337] [Remote→Local] 1

2025-05-07 14:54:20.773 [error] http: Error listing offerings: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": [
      "tools",
      1,
      "inputSchema"
    ],
    "message": "Required"
  }
]
```

I would like to know what I am doing wrong or if for some reason fastmcp is not compatible with mcp-remote. Thanks!

fauziralpiandi:
the error occurs because when a tool doesn't have parameters, FastMCP sends `inputSchema: undefined`, but `mcp-remote` expects `inputSchema` to always be an object (even if empty).

Solution:

we created a fix that changes FastMCP to always send an empty object (`{}`) instead of `undefined` for the `inputSchema` property when a tool has no parameters.

punkpeye:
[...]

PR #59, opened 2025-05-07
feat: configurable roots support

### LINKED ISSUE #37: List roots not supported" Error with FastMCP 1.21.0 in Programmatic Clients

fauziralpiandi:
Ability to explicitly disable roots functionality in server options, fixing compatibility issues with clients that don't support the roots capability.

- Resolve #37
- Ensure compatibility with #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ A TypeScript framework for building [MCP](https://glama.ai/mcp) servers capable 
 - [Prompt argument auto-completion](#prompt-argument-auto-completion)
 - [Sampling](#requestsampling)
 - [Configurable ping behavior](#configurable-ping-behavior)
-- Roots
+- [Roots](#roots-management)
 - CLI for [testing](#test-with-mcp-cli) and [debugging](#inspect-with-mcp-inspector)
 
 ## Installation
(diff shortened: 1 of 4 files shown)


punkpeye:
Thank you!

Issue #54, opened 2025-05-05
Add support for ajv as a validator implementation


btiernay:
It would be nice to support `ajv`-based validation so implementers can use OAS for metadata driven framework bootstrapping. This should be configured as a peer dependency (and probably others as well). As a consequence, this would aid https://github.com/punkpeye/fastmcp/issues/42

btiernay:
👋 @punkpeye curious to get your feedback to see if this would be a valuable addition. Thanks!

lucamorettibuilds:
I took a slightly different approach in #244 — instead of replacing the Zod-based validation, it adds a `jsonSchemaAdapter` that lets you pass plain JSON Schema objects as `schema` in tool definitions. Under the hood it uses `ajv` for validation but wraps results in the Standard Schema interface that FastMCP already expects.

This way existing Zod-based tools keep working unchanged, but teams that prefer raw JSON Schema (or are generating schemas programmatically) can use ajv without any breaking changes.

Happy to adjust the approach based on feedback — the key idea is treating ajv as an adapter rather than a replacement.

PR #60, opened 2025-05-07
feat: add support for `ajv` validation of parameters

btiernay:
### Add Support for JSON Schema (via AJV)

This PR adds support for using plain JSON Schema objects in the `parameters` field when defining tools—just like you can already do with Zod, ArkType, and Valibot.

### What’s new

You can now write tools like this using raw JSON Schema:

```ts
server.addTool({
  name: "add",
  description: "Add two numbers",
  parameters: {
    type: "object",
    properties: {
      a: { type: "number" },
      b: { type: "number" },
    },
    required: ["a", "b"],
    additionalProperties: false,
  },
  execute: async (args) => String(args.a + args.b),
});
```

- Behind the scenes, this uses AJV for validation.
- AJV and `ajv-formats` are added as peer dependencies.
- The implementation plugs into the existing validation system without changing the overall shape of how things work.
- Usage is optional and won’t impact folks sticking with Zod or others.

### Under the hood

- Detects JSON Schema objects with a type guard (`isJsonSchema`)
- Wraps them in a simple adapter so they look like a `StandardSchemaV1` validator internally
- Loads AJV dynamically, so it only kicks in if you’re actually using JSON Schema
- Keeps type safety intact, including through the dynamic import
- Errors from AJV are converted into the same format used elsewhere in FastMCP

### Other updates

- Added tests to confirm behavior and edge cases
- Updated the docs with examples
- Implementation stays lightweight—no new files or folders, just fits into the current structure

This is one step toward better support for OpenAPI-style workflows (see #54), and should make it easier to bring your own schemas without needing to convert them into validator-specific formats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,9 @@ const client = new Client(
   },
 );
 
-const transport = new StreamableHTTPClientTransport(new URL(`http://localhost:8080/stream`));
+const transport = new StreamableHTTPClientTransport(
+  new URL(`http://localhost:8080/stream`),
+);
 
 await client.connect(transport);
 ```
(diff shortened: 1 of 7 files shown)


punkpeye:
Closing this as it hasn't received a lot of attention/demand, and now the branch has drifted.

PR #80, opened 2025-05-25
fix(docs): update README to use await for async functions in examples

bianbianzhu:
Fix missing await when calling async function imageContent in the document 

### Reference
The type definition of `imageContent`:
```
declare const imageContent: (input: {
    buffer: Buffer;
} | {
    path: string;
} | {
    url: string;
}) => Promise<ImageContent>;
```

The type definition of `Content` (item of content list):
```
type Content = AudioContent | ImageContent | TextContent;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,10 +587,10 @@ server.addTool({
 
   // or...
   // execute: async (args) => {
-  //   const imgContent = imageContent({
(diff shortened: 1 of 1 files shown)


punkpeye:
Thank you

PR #84, opened 2025-05-28
fix: enchanted error messages

fauziralpiandi:
significantly improves error handling that make debugging faster and less frustrating ~


punkpeye:
Thank you!

PR #91, opened 2025-06-05
fix: HTTP Stream Transport Buffering

### LINKED ISSUE #90: ReportProgress Delayed in HTTP Stream Transport Due to Buffering (Not Fired Immediately)
## Description

When using the `reportProgress` function in a FastMCP tool (with HTTP stream transport), progress notifications are **not** sent immediately after the function call. Instead, they only appear **after the next `await` statement**, which is not the expected behavior.

This does **not** occur when using the `stdio` transport.

---

## ❗️Observed Behavior

In the following code snippet:

```tsx
// inside the addTool `execute` callback function
try {
  reportProgress({ progress: 0, total: 100 });
  await sleep(1000);

  reportProgress({ progress: 50, total: 100 });
  await sleep(1000);

  reportProgress({ progress: 90, total: 100 });
  await sleep(1000);

  reportProgress({ progress: 100, total: 100 }); // <- NOT WORKING in HTTP Stream transport

  return {
    content: [
      {
        type: "text",
        text: `Download from ${url} completed: This is a story about a cat`,
      },
    ],
    isError: false,
  };
}
```

- Each call to `reportProgress()` **only emits output after the subsequent `await sleep()` completes**, suggesting a delay.
- notification/progress for progress 100 will never be received.

---

## 🧪 Diagnostics

- Adding `await` before each `reportProgress()` **did not resolve** the issue.
- This behavior only appears when using **HTTP stream transport**. The same logic works fine with `stdio`.

---

## 💡 Potential Root Cause: HTTP Streaming and Buffering

This is **not** a JavaScript issue. It's due to **HTTP response buffering**:

My assumption (God bless Claude 4)
- FastMCP uses `mcp-proxy` with HTTP stream transport.
- Node.js HTTP servers, proxies, and clients often **buffer small writes** for performance.
- `await sleep()` gives the event loop time to **flush buffered responses**.

---

## ✅ Verification

Confirmed that the issue **does not occur** with `stdio` transport, which implies it's a **transport-level buffering issue**, not an application logic problem.

---

### ✅ Option 1: Explicit Flush Support

Extend the `StreamableHTTPServerTransport` with a `flush()` method.

1. Go to [src/server/streamableHttp.ts](https://github.com/modelcontextprotocol/typescript-sdk/compare/main...bianbianzhu:typescript-sdk:fix/add-flush-method-to-streamable-http-server-transport?expand=1#diff-fa2d8ab461bf11229eaa80d28cfa765b7244ab37a737a0b67f26aae8102175ee) of the official modelContextProtocol repo and add the following `flush` method.
```ts
flush() {
  this._streamMapping.forEach((response) => {
    try {
      const socket = response.socket;
      if (socket && typeof socket.flush === 'function') {
        socket.flush();
      } else if (socket && !socket.destroyed) {
        socket.uncork();
        socket.cork();
        socket.uncork();
      }

      if (response.chunkedEncoding && response.writable && !response.destroyed) {
        response.write('');
      }
    } catch (error) {
      console.warn('Warning: Failed to flush response stream:', error?.message ?? String(error));
    }
  });
}
```

Then call it like this after sending progress:

```ts
const reportProgress = async (progress) => {
  await this.#server.notification({
    method: "notifications/progress", 
    params: { ...progress, progressToken }
  });
 
  if (this.#server.transport?.flush) {
    transport.flush();
  }
};
```
🤦🏻‍♂️ Not sure if the official sdk wants to add this method or not.
---

### ✅ Option 2: Add Minimal Delay to Force Flush

This workaround forces a flush in the event loop:

```ts
await reportProgress({ progress: 50, total: 100 });
await new Promise((resolve) => setImmediate(resolve));
```
---

### ✅ Option 2 PLUS: Abstracted Progress Helper

Encapsulate flush delay into a helper:

```ts
async function flushProgress(process: Progress) {
  await reportProgress(process);
  await new Promise((resolve) => setImmediate(resolve));
}
```

Then use it like this:

```ts
await flushProgress({ progress: 55, total: 100 }); // must call with await
```

---

Let us know if you'd like a PR with `flush()` support added to the HTTP transport class.

fauziralpiandi:
When using FastMCP with HTTP Stream transport, progress notifications (`reportProgress`) and streaming content (`streamContent`) were not sent immediately to clients. Instead, they were buffered by the HTTP transport layer and only flushed when the next async operation occurred. So, we're added automatic flush behavior to both `reportProgress` and `streamContent` functions by inserting a minimal async operation after each notification:

```ts
// Force flush for HTTP stream transport to prevent buffering delays
await new Promise((resolve) => setImmediate(resolve));
```

- `setImmediate` schedules the callback for the next iteration of the event loop
- This gives Node.js time to flush any buffered HTTP responses
- The operation is lightweight and doesn't introduce significant delay
- It's transport-agnostic and works for both HTTP Stream and stdio

CLOSE #90

### DIFF
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -514,6 +514,76 @@ test("tracks tool progress", async () => {
   });
 });
 
+test("reports multiple progress updates without buffering (fixes issue #90)", async () => {
+  await runWithTestServer({
+    run: async ({ client }) => {
+      const progressCalls: Array<{ progress: number; total: number }> = [];
+      const onProgress = vi.fn((data) => {
+        progressCalls.push(data);
+      });
+
+      await client.callTool(
+        {
+          arguments: {
+            steps: 3,
+          },
+          name: "progress-test",
+        },
+        undefined,
+        {
+          onprogress: onProgress,
+        },
+      );
+
+      expect(onProgress).toHaveBeenCalledTimes(4);
+      expect(progressCalls).toEqual([
+        { progress: 0, total: 100 },
+        { progress: 50, total: 100 },
+        { progress: 90, total: 100 },
+        { progress: 100, total: 100 }, // This was previously lost due to buffering
+      ]);
+    },
+    server: async () => {
+      const server = new FastMCP({
+        name: "Test",
+        version: "1.0.0",
+      });
+
+      server.addTool({
+        description: "Test tool for progress buffering fix",
+        execute: async (args, { reportProgress }) => {
+          const { steps } = args;
+
+          // Initial
+          await reportProgress({ progress: 0, total: 100 });
+
+          for (let i = 1; i <= steps; i++) {
+            await delay(50); // Small delay to simulate work
+
+            if (i === 1) {
+              await reportProgress({ progress: 50, total: 100 });
+            } else if (i === 2) {
+              await reportProgress({ progress: 90, total: 100 });
+            }
+          }
+
+          // This was the critical test case that failed before the fix
+          // because there's no await after it, causing it to be buffered
+          await reportProgress({ progress: 100, total: 100 });
+
+          return "Progress test completed";
+        },
+        name: "progress-test",
+        parameters: z.object({
+          steps: z.number(),
+        }),
+      });
+
+      return server;
+    },
+  });
+});
+
 test("sets logging levels", async () => {
   await runWithTestServer({
     run: async ({ client, session }) => {
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -514,7 +514,7 @@ test("tracks tool progress", async () => {
   });
 });
 
-test("reports multiple progress updates without buffering (fixes issue #90)", async () => {
+test("reports multiple progress updates without buffering", async () => {
   await runWithTestServer({
     run: async ({ client }) => {
       const progressCalls: Array<{ progress: number; total: number }> = [];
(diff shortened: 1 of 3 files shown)


bianbianzhu:
@fauziralpiandi Thanks for the rapid fix! 🔥🔥🔥

I have a suggestion on this:
Do you think we need to first check if the transport fails into SSE or HTTP stream, then use the event loop behaviour to flush the buffer plz? For stdio, we don't really have any issues. 
Something like:
```typescript
 const reportProgress = async (progress: Progress) => {
          await this.#server.notification({
            method: "notifications/progress",
            params: {
              ...progress,
              progressToken,
            },
          });

          function isStreamableHTTPorSSE(transport: Transport | undefined) {
            const transportName = transport?.constructor.name;

            return (
              transportName === "SSEServerTransport" ||
              transportName === "StreamableHTTPServerTransport"
            );
          }

          const transport = this.#server.transport;
          if (isStreamableHTTPorSSE(transport)) {
            await new Promise((resolve) => setImmediate(resolve));
          }
        };
     
     I also realise, although `SSEServerTransport` and `StreamableHTTPServerTransport` can be imported directly from @modelcontextprotocal/sdk, using `transport instanceof <transport class name>` will always give us false. 
     Looks like we have a module duplication issue here, and have to use `constructor.name` instead. 
     
 Keen to know your thoughts on this. 
 Thanks

PR #94, opened 2025-06-06
perf: optimize event loop flushing for different transport types

fauziralpiandi:


bianbianzhu:
Thanks! This implementation is elegant. 🔥

punkpeye:
Thank you!

Issue #103, opened 2025-06-10
onConnect function should not be async


bianbianzhu:
## Background
In `FastMCP.ts` -> `FastMCP` class -> `start()`
If the transport is "httpStream",  we invoke `startHTTPServer` from `mcp-proxy` library and setup the onConnect event handler.
The onConnect function is async

```typescript
onConnect: async (session) => {
  this.#sessions.push(session);
  this.emit("connect", { session });
},
```

However, when this `onConnect` gets invoked in `startHttpServer.ts`, it's called without `await`.
```typescript

if (!sessionId && isInitializeRequest(body)) {
  // ...omit logic
  try {
          server = await createServer(req);
        } catch (error) {
          if (error instanceof Response) {
            res.writeHead(error.status).end(error.statusText);
            return true;
          }
          res.writeHead(500).end("Error creating server");
          return true;
        }
        server.connect(transport);
        onConnect?.(server); // No await!
  // ...omit logic
}
```

## Problem
- **Unhandled Promise**: The async function returns a Promise that's ignored, potentially causing unhandled promise rejection warnings
- **Race Conditions**: If onConnect ever does async work (like waiting for session initialization), the calling code continues without waiting for completion
- **Error Handling**: Any errors thrown in the async onConnect won't be caught by the calling code

## To prove
Use logging to prove that onConnect is not properly awaited

1. In `FastMCP.ts` -> `FastMCP` class -> `start()` -> `startHTTPServer()` -> `onConnect`, add the following loggoings.
```typescript
        onConnect: async (session) => {
          console.log(`[FastMCP DEBUG] onConnect START - ${new Date().toISOString()}`);
          
          this.#sessions.push(session);

          // Add a small delay to make the async nature more obvious
          await new Promise(resolve => setTimeout(resolve, 100));
          console.log(`[FastMCP DEBUG] After 100ms delay`);

          this.emit("connect", {
            session,
          });
          
          console.log(`[FastMCP DEBUG] onConnect END - ${new Date().toISOString()}`);
        },

```
2. In `FastMCP.test.ts` -> test case `"HTTP Stream: calls a tool"`, add loggings before and after `client.connect(transport)`
```typescript
    // create server
    // add tools to server
    // start server
    // instantiate client transport (http stream)
    // create session promise
    console.log(
      `[TEST DEBUG] About to call client.connect() - ${new Date().toISOString()}`
    );
    await client.connect(transport); // client connects to the client transport
    console.log(
[...]

PR #104, opened 2025-06-10
fix: change onConnect to synchronous function in FastMCP

bianbianzhu:
## Related issue
This fix is relevant to issue #103 
### Add `await` before each `onConnect` in `startHTTPServer.ts` in `mcp-proxy`
```typescript
// For example, in startHTTPServer.ts line 128
await onConnect?.(server);
```
However, this requires a PR to `mcp-proxy` 🤦🏻‍♂️.

In `fastmcp`, currently we don't have async operations in `onConnect` handler:
```typescript
 onConnect: (session) => {
          this.#sessions.push(session);

          this.emit("connect", {
            session,
          });
        },
 ```
 It is more reasonable to simply remove `async` keywords to avoid any confusions. 
 
 ❓ Not sure if we need to introduce any test cases for this.

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -1718,7 +1718,7 @@ export class FastMCP<
             session,
           });
         },
-        onConnect: async (session) => {
+        onConnect: (session) => {
           this.#sessions.push(session);
 
           this.emit("connect", {


punkpeye:
Hey, thanks for documenting the issue and raising a PR.

However, this is a bug in mcp-proxy – those methods should have been awaited.

Fixed in v5:

https://github.com/punkpeye/mcp-proxy/releases/tag/v5.0.0

Would you be up to updating FastMCP to match these changes?

PR #105, opened 2025-06-10
Fix/upgrade mcp proxy and add async to on close

bianbianzhu:
## Background
Please check this issue #103 
As mentioned by @punkpeye, "this is a bug in mcp-proxy – those methods should have been awaited."

This has been fixed in v5 of mcp-proxy:
https://github.com/punkpeye/mcp-proxy/releases/tag/v5.0.0

## What has this PR done
1. Update the `mcp-proxy` to latest version 5.0.0
2. Add `async` keywords to the `onClose` handler of the `startHTTPServer` function (`onConnect` is already async)
3. Update the test case to use mockResolvedValue to match the async nature of the handlers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "execa": "^9.6.0",
     "file-type": "^21.0.0",
     "fuse.js": "^7.1.0",
-    "mcp-proxy": "^4.0.0",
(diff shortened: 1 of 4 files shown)


punkpeye:
Thank you!

PR #116, opened 2025-06-15
fix(test): update progress steps in FastMCP test case

OR13:


bianbianzhu:
I think there's something wrong with the test case here. It is flaky. I saw onProgress got called both 3 times and 4 times in different runs, and this is probably the reason why this test case passed.

But `steps` should be 4

punkpeye:
Maybe add retry instead?

```diff
- test("reports multiple progress updates without buffering", async () => {
+ test("reports multiple progress updates without buffering", { retry: 3 }, async () => {
```

Issue #118, opened 2025-06-16
[Bug] ERR_PACKAGE_PATH_NOT_EXPORTED: file-type/package.json when using fastmcp@3.x with tsx/ESM


codeByJiaDong:
### Describe the bug

When running a project with `fastmcp@3.1.1` (or latest 3.x) using `tsx` or in a Node.js ESM environment, I get the following error:

This happens both when running with `npx tsx src/mcp-server/index.ts` and when using `concurrently` to start Next.js and MCP together.

### To Reproduce

1. Create a new project and install fastmcp:
   ```
   pnpm add fastmcp@3
   ```
2. Create a simple TypeScript entry file that imports and starts FastMCP.
3. Run the file with:
   ```
   npx tsx src/mcp-server/index.ts
   ```
4. See the error above.

### Expected behavior

The MCP server should start normally in ESM/tsx environments.

### Environment

- Node.js: v20.9.0
- fastmcp: 3.1.1
- file-type: (as installed by fastmcp)
- OS: macOS (Darwin 24.3.0)
- tsx: 4.20.3
- pnpm: 8.x

### Additional context

- Downgrading to `fastmcp@2` works fine.
- This seems to be caused by `file-type` not exporting a main entry in its package.json, which breaks ESM/tsx compatibility.
- See also: [Node.js ERR_PACKAGE_PATH_NOT_EXPORTED documentation](https://nodejs.org/api/packages.html#exports)

Please consider updating the file-type dependency or adjusting the import logic for better ESM compatibility.

Thank you!

codeByJiaDong:
npx tsx src/mcp-server/index.ts
node:internal/errors:497
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/aiden/work/front/info-element/node_modules/.pnpm/fastmcp@3.1.1/node_modules/file-type/package.json
    at __node_internal_captureLargerStackTrace (node:internal/errors:497:5)
    at new NodeError (node:internal/errors:406:5)
    at exportsNotFound (node:internal/modules/esm/resolve:268:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:542:13)
    at resolveExports (node:internal/modules/cjs/loader:547:36)
    at Module._findPath (node:internal/modules/cjs/loader:621:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1034:27)
[...]

PR #119, opened 2025-06-16
fix: replaced static imports with dynamic imports

fauziralpiandi:
`ERR_PACKAGE_PATH_NOT_EXPORTED` error when using `fastmcp@3.x` with tsx/ESM due to `file-type` package's ESM-only exports.

the fix resolves issue #118 by ensuring `file-type` is loaded dynamically at runtime rather than during module parsing.


fauziralpiandi (commented):
```sh
node -e "import('./dist/FastMCP.js').then(async ({ imageContent }) => { await imageContent({ buffer: Buffer.from('test') }); console.log('✓'); })"
```

Issue #122, opened 2025-06-17
Extend smaplingRequest to support passing options


OrenMe:
Currently there is no way to pass options to `requestSampling`
Add options in function call to allow extending it
https://github.com/punkpeye/fastmcp/blob/e30912374028710dc77014b608512189e00b561d/src/FastMCP.ts#L878-L881

fauziralpiandi:
implemented! thanks @OrenMe

PR #125, opened 2025-06-17
feat: add support for passing options to requestSampling

OrenMe:
Solves https://github.com/punkpeye/fastmcp/issues/122

Note: for some reason I wasn't able to import the RequestOptions from the protocol file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1368,26 +1368,27 @@ await session.requestSampling(
     onprogress: (progress) => {
       console.log(`Progress: ${progress.progress}/${progress.total}`);
     },
-    
+
     // Abort signal for cancelling the request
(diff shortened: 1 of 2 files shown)


punkpeye:
@OrenMe just need help passing linting

OrenMe:
Hi @punkpeye , thank u for reviewing
I pushed the fix and returned usage of type system
There is inherit issue with the return type due to mismatch between payload of content, TextContent has text filed while AudioContent and ImageContent doesn't have it, but this is not related to my change

Also ran the linter and prettier which moved the SamplingResponse

OrenMe:
Thank u! Love the project

PR #130, opened 2025-06-24
feat: Support Tool calls returning ResourceLink object.

pattobrien:
Adds support for MCP ResourceLink objects.


punkpeye:
Thank you

PR #131, opened 2025-06-25
Allow configuring a custom eventStore

Floriferous:
`eventStore` can't be customized when using FastMCP, this allows passing one to `startHTTPServer`.

Maybe there are other options that should be passed along for full customizability?


punkpeye:
Thank you for your contribution. Please ensure that all checks are passing.

PR #151, opened 2025-07-17
feat: allow custom friendlyErrors message on parameters validation error

soamati:
### Improve Validation Error Messaging via Custom Formatter

In some cases, validation errors are too vague to generate helpful responses for users.
For example, when using Zod schemas with union types, the default error message is often just "Invalid input", which lacks useful detail.

This PR introduces a new utils field to the server constructor, currently supporting a formatInvalidParamsErrorMessage callback. This allows developers to intercept and customize the validation error message based on the raw Zod issues.

By enabling this customization, agents can better understand what went wrong and return clearer, more informative error messages to users.

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -670,6 +670,14 @@ type ServerOptions<T extends FastMCPSessionAuth> = {
      */
     enabled?: boolean;
   };
+  /**
+   * General utilities
+   */
+  utils?: {
+    formatInvalidParamsErrorMessage?: (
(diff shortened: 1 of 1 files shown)


punkpeye:
Looks good, but please add tests.

soamati:
> Looks good, but please add tests.

Great! Test added

punkpeye:
Thank you – do you think there is a way to generalize "a good error", or should we assume that most will need to customize this?

PR #175, opened 2025-09-12
feat: add request.params._meta to Context

matiasanaya:
Not sure if this is the right approach. It is the simplest I could come up with right now. Though what I would probably prefer is something [like a middleware](https://gofastmcp.com/servers/middleware), since `context` is only available for tools but `_meta` is everywhere. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,7 +771,53 @@ The `log` object has the following methods:
 
 #### Errors
 
-The errors that are meant to be shown to the user should be thrown as `UserError` instances:
+FastMCP supports two ways to handle errors in tool execution:
+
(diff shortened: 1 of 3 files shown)


matiasanaya:
> What were you trying to achieve?

I'm trying to read [_meta fields](https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta) from within a tool execution. This I need because I want to build an x402 enabled MCP server, and per [the spec](https://github.com/coinbase/x402/blob/main/specs/transports/mcp.md) the x402 data is sent in the `_meta` fields.

PR #199, opened 2025-10-25
Add support for runtime injection of  tools/prompts/resources

subbu963:
### Problem:
Existing sessions doesnt get the latest tools, etc when they get updated.

### Fix:
This PR fixes it by adding support of dynamic injection and also triggers `list_changed` notifications for them

Test cases added.

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -1249,10 +1249,18 @@ export class FastMCPSession<
     this.triggerListChangedNotification("notifications/tools/list_changed");
   }
 
-  triggerListChangedNotification(method: string) {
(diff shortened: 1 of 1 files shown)


punkpeye:
I am aligned with the need for dynamic updates, but why add all the helpers for batching changes?

`addTools`, etc. is not necessary when there is `addTool`

subbu963:
> > client will be overwhelmed by the list_changed updates.
> 
> Why would the client be 'overwhelmed'?

let say i am doing 50 tool modifications post the `start()`, there will be 50 updates instead of one.

Issue #202, opened 2025-10-29
Add CORS options


OlivierCuyp:
I'm using docker to create my MCP Servers and using MCP Inspector to test them locally.
Doing so I have a CORS issue on the MCP Inspector and I have no way to configure the CORS.
It would nice to have more control over the CORS.

punkpeye:
Implemented in v4.3.0 via #266

PR #203, opened 2025-10-29
feat: add CORS configuration support

OlivierCuyp:
Proposal for #202

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "execa": "^9.6.0",
     "file-type": "^21.0.0",
     "fuse.js": "^7.1.0",
-    "mcp-proxy": "^5.8.1",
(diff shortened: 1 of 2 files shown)


src/FastMCP.test.ts
```diff
@@ -4387,3 +4387,250 @@ test("tools can access client info", async () => {
     },
   });
 });
+
+test(
+  "CORS configuration: default CORS headers",
+  { timeout: 20000 },
+  async () => {
+    const port = await getRandomPort();
```
punkpeye: Is there a reason not to use `runWithTestServer` helper that the rest of the tests are using?

punkpeye (changes requested):

PR #209, opened 2025-11-16
fix: add completions capability for Cursor compatibility

Justiceleeg:
## Description
Adds support for the completions capability required by Cursor. Previously, FastMCP would throw errors when receiving completion requests it didn't support, causing Cursor to fail when starting MCP servers.

## Changes
- Added `completions` capability to server capabilities
- Updated completion handler to return empty completions instead of throwing errors for unsupported types
- This allows Cursor to work with FastMCP servers even if they don't implement all completion types

## Testing
- Tested with Cursor and confirmed the server starts successfully
- Verified that completion requests return empty arrays when not supported

## Related Issues
Fixes: [eyaltoledano/claude-task-master#1413](https://github.com/eyaltoledano/claude-task-master/issues/1413)

* this is the first time I've contributed to OSS be nice to me

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -1404,13 +1404,16 @@ export class FastMCPSession<
           (prompt) => prompt.name === request.params.ref.name,
         );
 
-        if (!prompt || !prompt.complete) {
-          // Return empty completion instead of throwing error
-          return {
-            completion: {
-              values: [],
-            },
-          };
+        if (!prompt) {
+          throw new UnexpectedStateError("Unknown prompt", {
+            request,
+          });
+        }
+
+        if (!prompt.complete) {
+          throw new UnexpectedStateError("Prompt does not support completion", {
+            request,
+          });
         }
 
         const completion = CompletionZodSchema.parse(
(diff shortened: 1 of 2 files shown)


punkpeye:
Hey, can you add a test showing what this fixes?

Crunchyman-ralph:
@punkpeye I thought this was resolved, but cursor is complaining about this, here are some issues we've been having from the community:
- https://github.com/eyaltoledano/claude-task-master/issues/1411
- https://github.com/eyaltoledano/claude-task-master/issues/1404
- https://github.com/eyaltoledano/claude-task-master/issues/1405

src/FastMCP.ts
```diff
@@ -1402,16 +1404,13 @@ export class FastMCPSession<
...
-          throw new UnexpectedStateError("Prompt does not support completion", {
-            request,
-          });
+        if (!prompt || !prompt.complete) {
+          // Return empty completion instead of throwing error
+          return {
+            completion: {
+              values: [],
+            },
+          };
```
Crunchyman-ralph: I don't know if this makes sense or if it is even relevant to solving your issue...

I believe there is a difference between completions, and prompt completions...

Though I haven't dug too deep in the mcp spec.

Crunchyman-ralph (changes requested):
reviewed, have a couple of questions.

Though I'm not a maintainer, but really need this out, so I'll push as hard as possible lol

PR #223, opened 2026-01-14
refactor: use `Map` for internal `prompts` and `resources` storage

fauziralpiandi:
migrated `prompts`, `resources`, and `resourceTemplates` from `Array` to `Map` for better performance and data integrity.

**why?**

- **automatic deduplication:** using `Map` ensures that prompts (by `name`) and resources (by `uri`) are unique. any subsequent additions with the same key will overwrite the old entry instead of creating duplicates in an array.
- **improved performance:** lookups for specific prompts or resources (e.g., `getPrompt`, `readResource`) are now **O(1)** instead of **O(n)**, as we no longer need to iterate through an array to find a match.
- **consistent state:** internal methods now manage state more cleanly using `.clear()` and `.set()`, ensuring that the server always serves the most up-to-date definitions.

### DIFF
--- a/src/FastMCP.map-storage.test.ts
+++ b/src/FastMCP.map-storage.test.ts
@@ -1,0 +1,120 @@
+import { expect, test, vi } from "vitest";
+
+import { FastMCPSession } from "./FastMCP.js";
+
+const mockLogger = {
+  debug: vi.fn(),
+  error: vi.fn(),
+  info: vi.fn(),
+  log: vi.fn(),
+  warn: vi.fn(),
+};
+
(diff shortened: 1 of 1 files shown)


fauziralpiandi:
```txt
RUN  v4.0.15 /home/runner/workspace

✓ src/FastMCP.map-storage.test.ts (3 tests)
  ✓ FastMCPSession internal storage prevents duplicates for prompts
  ✓ FastMCPSession internal storage prevents duplicates for resources
  ✓ FastMCPSession handles list changes correctly by clearing old state

Test Files 1 passed (1)
  Tests 3 passed (3)
```

PR #226, opened 2026-01-18
fix: avoid unhandled rejection in tool timeout

### LINKED ISSUE #225: httpStream stateless: UserError + timeoutMs causes unhandled rejection and exits process
## Summary
When a tool throws `UserError` and `timeoutMs` is set, FastMCP returns a proper `isError` response but the Node process still exits due to an unhandled rejection. This happens because `executeToolPromise.finally(...)` creates a rejecting promise that is never handled.

## Environment
- fastmcp: 3.26.9
- Node: v24.13.0 (also reproducible on v22 with `NODE_OPTIONS=--unhandled-rejections=strict`)
- Transport: `httpStream` (stateless)

## Minimal repro
```js
import { FastMCP, UserError } from "fastmcp";

const server = new FastMCP({
  name: "fastmcp-usererror-repro",
  version: "0.0.0",
  instructions: "Repro: tool throws UserError over httpStream."
});

server.addTool({
  name: "throw_user_error",
  description: "Always throws a UserError.",
  timeoutMs: 1000,
  execute: async () => {
    throw new UserError("Boom: should be returned, not crash.");
  }
});

await server.start({
  transportType: "httpStream",
  httpStream: { port: 8081, stateless: true }
});
```
Request:
```bash
curl -sS -D - \
  -H 'accept: application/json, text/event-stream' \
  -H 'content-type: application/json' \
  -H 'mcp-protocol-version: 2025-03-26' \
  -X POST http://localhost:8081/mcp \
  --data '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"throw_user_error"}}'
```

## Expected
- Response with `isError: true`
- Server keeps running

## Actual
- Response is returned
- Process exits with uncaught `UserError` (unhandled rejection)

## Root cause
In `FastMCP.ts`, the timeout branch does:
```ts
executeToolPromise.finally(() => clearTimeout(timeoutId));
```
That `finally()` returns a promise which rejects when the tool throws. Because it is not awaited or handled, Node treats it as an unhandled rejection and exits.

## Proposed fix
Use a non-throwing handler to clear the timeout:
```ts
executeToolPromise.then(
  () => clearTimeout(timeoutId),
  () => clearTimeout(timeoutId)
);
```
(Alternative: `void executeToolPromise.catch(() => {});` after the race, but the `then` change is sufficient.)

Happy to provide a PR if you want.

Darley-Wey:
## Summary
- Avoids creating an unhandled rejection when a tool throws and `timeoutMs` is enabled.
- Replaces `executeToolPromise.finally(...)` with a non-throwing `then(..., ...)` cleanup.

## Context
This fixes the crash described in #225 where `UserError` is returned correctly but the process exits due to an unhandled rejection.

## Testing
- Not run (behavioral change only).

Fixes #225

### DIFF
--- a/src/FastMCP.ts
+++ b/src/FastMCP.ts
@@ -1971,10 +1971,10 @@ export class FastMCPSession<
                 }, tool.timeoutMs);
 
                 // If promise resolves first
-            executeToolPromise.then(
-              () => clearTimeout(timeoutId),
-              () => clearTimeout(timeoutId),
-            );
+                executeToolPromise.then(
+                  () => clearTimeout(timeoutId),
+                  () => clearTimeout(timeoutId),
+                );
               }),
             ])
           : executeToolPromise)) as


Darley-Wey:
Added a regression test covering `UserError` + `timeoutMs` (ensures no unhandled rejection is emitted). Still not running tests locally; the change is in `src/FastMCP.test.ts`.

Darley-Wey:
Follow-up: `pnpm lint` passes locally (prettier/eslint/tsc/jsr publish dry-run).

Issue #42, opened 2025-04-23
Please implement OpenAPI to MCP converter in the TypeScript version


Barakaa5:
Hi! Thanks for the great work on fastmcp-typescript.

It would be amazing to have feature parity with the Python version by supporting OpenAPI-to-MCP conversion.

The Python package currently supports generating MCP servers directly from OpenAPI specs:

```
from fastmcp import FastMCP
import httpx

openapi_spec = {...}  # OpenAPI spec dict
http_client = httpx.AsyncClient(base_url="https://api.yourservice.com")
mcp_server = FastMCP.from_openapi(openapi_spec, client=http_client)
mcp_server.run()
The rules followed:

```
GET -> MCP resources

GET with path params -> MCP resource templates

Other HTTP methods -> MCP tools

It would be great to have the same fromOpenAPI() method in TypeScript. Would love to collaborate or test if needed!

Thanks 🙌

punkpeye:
Would you be open to contributing this change?

RJL667:
Two PRs have now closed on this (#193, #243), and the closing note on #243 was
"needs a lot more work" without an itemised list. I've built and shipped this
exact converter and benchmarked it against seven production specs, so here is
what "a lot more work" concretely consists of. Posting it because the next person
who attempts this deserves the list up front rather than after five months of
review.

**1. External `$ref` splicing — this is the one that decides whether it works at all.**
Serious APIs do not ship one file. They keep one YAML per URL and assemble at
build time, so the entry document is a few KB of pure `$ref` and the operations
live in `resources/<tag>/*.yaml`. A resolver that only handles local `#/...`
pointers sees **zero operations** and the spec reads as empty rather than as
unsupported. You need relative paths, absolute URLs, and the
`other.yaml#/fragment` form.

The sharp edge underneath it: **an external `$ref` resolves relative to the
document it came from.** If you download the spec to a temp file first and then
parse, every relative ref resolves against the wrong base and you get the same
silent zero. The spec URL has to stay the resolution base.

**2. A relative `servers[0].url` will pass every schema test and fail every real call.**
OpenAPI permits `"url": "/api/v3"` (Petstore does exactly this). Emit it verbatim
and the generated client raises `unknown url type: '/api/v3/store/inventory'` on
the first request — while `tools/list` and JSON-Schema validation stay green.
Resolve it against the spec document's own origin. I shipped this bug, and the
schema-level test suite did not notice.

[...]

PR #244, opened 2026-02-16
feat: add jsonSchemaAdapter for plain JSON Schema support — closes #54

### LINKED ISSUE #54: Add support for ajv as a validator implementation

lucamorettibuilds:
## Summary

Closes #54

Adds a `jsonSchemaAdapter()` utility that wraps plain JSON Schema objects with a [Standard Schema v1](https://github.com/standard-schema/standard-schema) interface. This lets users define tool parameters using raw JSON Schema — no Zod, Yup, or other validation library required.

## How it works

1. **`jsonSchemaAdapter(schema)`** returns a `StandardSchemaV1`-compatible object
2. **Runtime validation** uses [AJV](https://ajv.js.org/) (lazy-loaded, optional peer dep)
3. **Fast-path**: The adapter exposes `__jsonSchema` so FastMCP can skip the `toJsonSchema()` conversion and use the raw schema directly — zero overhead
4. **`ajv-formats`** is optionally detected for format validation (e.g. `email`, `uri`)

## Usage

```ts
import { FastMCP, jsonSchemaAdapter } from "fastmcp";

const server = new FastMCP({ name: "Example" });

server.addTool({
  name: "greet",
  description: "Greet a user",
  parameters: jsonSchemaAdapter({
    type: "object",
    properties: {
      name: { type: "string" },
    },
    required: ["name"],
  }),
  execute: async ({ name }) => `Hello, ${name}!`,
});
```

## Changes

| File | What |
|------|------|
| `src/jsonSchemaAdapter.ts` | The adapter function + `JsonSchemaObject` type |
| `src/jsonSchemaAdapter.test.ts` | 5 tests: valid input, invalid input, `__jsonSchema` fast-path, vendor metadata, nested objects |
| `src/FastMCP.ts` | Re-export from barrel + fast-path detection in `inputSchema` resolution |

## Design decisions

- **AJV is a peer dep, not a direct dep** — keeps the fastmcp bundle lightweight. Users who want JSON Schema support install `ajv` themselves. The adapter throws a clear error message if AJV is missing.
- **`__jsonSchema` fast-path** — avoids the `toJsonSchema()` roundtrip since we already have valid JSON Schema. FastMCP checks for this property before falling back to the standard conversion.
- **No changes to existing behavior** — all existing Zod/Valibot/ArkType schemas continue to work exactly as before.

## Related

- Issue: #54
- Requested by multiple users who want to use JSON Schema directly from OpenAPI specs or LLM-generated schemas

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ A TypeScript framework for building [MCP](https://glama.ai/mcp) servers capable 
 - [Typed server events](#typed-server-events)
 - [Prompt argument auto-completion](#prompt-argument-auto-completion)
 - [Sampling](#requestsampling)
+- [Elicitation](#elicitation)
 - [Configurable ping behavior](#configurable-ping-behavior)
 - [Health-check endpoint](#health-check-endpoint)
 - [Roots](#roots-management)
@@ -139,6 +140,8 @@ This will start the server and listen for HTTP streaming connections on `http://
 
 > **Note:** You can also customize the endpoint path using the `httpStream.endpoint` option (default is `/mcp`).
 
+> **Note:** To serve HTTP streaming and built-in OAuth routes under an issuer path, set `httpStream.basePath` (for example, `/issuer1`). This exposes authorization server metadata at `/.well-known/oauth-authorization-server/issuer1` per RFC 8414.
+
 > **Note:** This also starts an SSE server on `http://localhost:8080/sse`.
 
 You can connect to these servers using the appropriate client transport.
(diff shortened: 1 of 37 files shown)


lucamorettibuilds:
Hey @punkpeye — CI is blocked on first-time contributor approval (`action_required`). Could you approve the workflow run when you get a chance? The implementation wraps ajv in a Standard Schema–compatible adapter so existing Zod tools continue working while plain JSON Schema objects can be passed as `schema` directly. Happy to adjust anything.

ymxlx:
I hit #54 myself and came here first rather than open a competing PR — this adapter is the right approach. Wrapping ajv in a Standard Schema adapter (instead of threading raw JSON Schema through the whole validate path) keeps it consistent with how FastMCP already treats Zod / ArkType / Valibot, and it's the missing piece for #42 (OpenAPI → MCP), since an OAS operation hands you a JSON Schema directly.

Looks like the only things left are the first-time-contributor CI approval and a rebase — the branch has drifted to conflicting since February. @punkpeye, worth kicking off the workflow run? And @lucamorettibuilds, if it helps I'm glad to test the adapter against a handful of real-world JSON Schemas (nested objects, `format` keywords, `$ref`) to shake out edge cases.

PR #254, opened 2026-03-18
feat: add onToolCall callback

### LINKED ISSUE #163: Custom Logger Swallowing Tool Call Logs
Hello! Thanks for your work on FastMCP!

I'm having some trouble with the recently released custom logger feature, and I suspect it's a bug.

My customer logger is getting picked up my the framework for its internal logging, but any logs from a tool call are getting swallowed.

Here's my `index.ts` file:

```typescript
import { FastMCP } from "@punkpeye/fastmcp";
import { z } from "@zod/zod";
import Logger from "@/logger.ts";

export function run() {
  const logger = new Logger();
  logger.info("Hello from the logger, outside of FastMCP.");
  const server = new FastMCP({
    logger: new Logger(),
    name: "My Server",
    version: "1.0.0",
  });

  server.addTool({
    name: "add",
    description: "Add two numbers",
    parameters: z.object({
      a: z.number(),
      b: z.number(),
    }),
    execute: async (args, { log }) => {
      log.info("This message is getting sent into a blackhole.");
      console.log("Proof the tool call is running.");
      return String(args.a + args.b);
    },
  });

  server.start({
    transportType: "httpStream",
    httpStream: {
      endpoint: "/sse",
      port: 3001,
    },
  });
}
```

And here is my logging implementation. I'm using logtape, and it's coming out lovely! Aside from this one small issue.

```typescript
import {
  configure,
  getConsoleSink,
  getLogger,
  Logger as TapeLogger,
} from "@logtape/logtape";
import { getPrettyFormatter } from "@logtape/pretty";
import { Logger } from "@punkpeye/fastmcp";

class PrettyLogger implements Logger {
  readonly logger: TapeLogger;

  constructor() {
    this.logger = getLogger("last.engineer");
  }

  debug(...args: unknown[]): void {
    const msg = args.shift();
    // If we've been given additional arguments, then we print them.
    // Otherwise, we omit them from the format string.
    const logline = args.length ? `${msg} {*}` : `${msg}`;

    this.logger.debug(logline, { ...args });
  }

  info(...args: unknown[]): void {
    console.log(
      "Proof that the info function is not called from the tool (this line will not appear after the tool call)",
    );
    const msg = args.shift();
    // If we've been given additional arguments, then we print them.
    // Otherwise, we omit them from the format string.
    const logline = args.length ? `${msg} {*}` : `${msg}`;
    this.logger.info(logline, { ...args });
  }

  log(...args: unknown[]): void {
    this.info(args);
  }

  warn(...args: unknown[]): void {
    const msg = args.shift();
    // If we've been given additional arguments, then we print them.
    // Otherwise, we omit them from the format string.
    const logline = args.length ? `${msg} {*}` : `${msg}`;
    this.logger.warn(logline, { ...args });
  }

  error(...args: unknown[]): void {
    const msg = args.shift();
    // If we've been given additional arguments, then we print them.
    // Otherwise, we omit them from the format string.
    const logline = args.length ? `${msg} {*}` : `${msg}`;
    this.logger.error(logline, { ...args });
  }
}

await configure({
  sinks: {
    console: getConsoleSink({
      formatter: getPrettyFormatter(),
    }),
  },
  loggers: [
    {
      category: "last.engineer",
      lowestLevel: "info",
      sinks: ["console"],
    },
    // Disable the metalogger. This is not really best for production, but it's not
    // clear how we would want to represent these meta-logs.
    { category: ["logtape", "meta"], sinks: [] },
  ],
});

export default PrettyLogger;
```

As you'll be able to see, the tool call runs successfully, and the `console.log`s show that there's nothing wrong with _e.g._ stdout. It's just that the `log` argument in the tool call context does not actually call `.info`. If it did, the `console.log` I have in there would fire. Since we can see `console.log` working from the tool call, but not from inside `log.info`, I have to conclude the issue is with the injected logger.

I can give a full repo to reproduce if you like; I'm running Deno, so it may not be a one-liner for you to reproduce since you'd need to set up the same import mapping as me. 

Thanks for your time and assistance.

LikiosSedo:
## Summary

- Add `onToolCall` callback option to `ServerOptions`, invoked when a tool is called
- Consumers can use it to log, audit, or track tool usage however they prefer

## Changes

- **`src/FastMCP.ts`**: Add `onToolCall` option to `ServerOptions`, pass it through to `FastMCPSession`, and invoke it before tool execution with `{ toolName, arguments }`
- **`src/FastMCP.test.ts`**: Add test verifying `onToolCall` is invoked with correct tool name and arguments

## Testing

- All tests pass
- Linting passes (Prettier + ESLint)
- Build succeeds

Fixes #163

### DIFF
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -837,6 +837,64 @@ test("sends logging messages to the client", async () => {
   });
 });
 
+test("tool log calls are forwarded to custom logger", async () => {
+  const customLogger = {
+    debug: vi.fn(),
+    error: vi.fn(),
+    info: vi.fn(),
+    log: vi.fn(),
+    warn: vi.fn(),
+  };
+
+  await runWithTestServer({
+    run: async ({ client }) => {
+      await client.callTool({
+        arguments: {
+          a: 1,
+          b: 2,
+        },
+        name: "add",
+      });
+
+      expect(customLogger.debug).toHaveBeenCalledWith("debug message", {
+        foo: "bar",
+      });
+      expect(customLogger.error).toHaveBeenCalledWith(
+        "error message",
+        undefined,
+      );
+      expect(customLogger.info).toHaveBeenCalledWith("info message", undefined);
+      expect(customLogger.warn).toHaveBeenCalledWith("warn message", undefined);
+    },
+    server: async () => {
+      const server = new FastMCP({
+        logger: customLogger,
+        name: "Test",
+        version: "1.0.0",
+      });
+
+      server.addTool({
+        description: "Add two numbers",
+        execute: async (args, { log }) => {
+          log.debug("debug message", { foo: "bar" });
+          log.error("error message");
+          log.info("info message");
+          log.warn("warn message");
+
+          return String(args.a + args.b);
+        },
+        name: "add",
+        parameters: z.object({
+          a: z.number(),
+          b: z.number(),
+        }),
+      });
+
+      return server;
+    },
+  });
+});
+
 test("adds resources", async () => {
   await runWithTestServer({
     run: async ({ client }) => {
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -837,14 +837,8 @@ test("sends logging messages to the client", async () => {
   });
 });
 
-test("tool log calls are forwarded to custom logger", async () => {
-  const customLogger = {
-    debug: vi.fn(),
-    error: vi.fn(),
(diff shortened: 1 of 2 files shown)


punkpeye:
How about just adding `onToolCall` callback? Then consumer can choose to log whatever they want

LikiosSedo:
Done — replaced log forwarding with an `onToolCall` callback on `ServerOptions`. Invoked before execution with `{ toolName, arguments }`.

Issue #201, opened 2025-10-29
OAuth Proxy


powrojsx:
Hey, does the Typescript implementation support OAuth Proxy simillar to Python https://gofastmcp.com/servers/auth/oauth-proxy? 

If not, are there plans to add it, or maybe someone already did that?

lucamorettibuilds:
If you are looking at this from the angle of injecting auth tokens into MCP tool calls without exposing them to the LLM, [janee](https://github.com/rsdouglas/janee) handles that use case — it sits as a proxy between the client and server, injecting secrets (OAuth tokens, API keys, etc.) into tool call parameters at the transport layer.

For the OAuth proxy pattern specifically, janee supports defining secret values that get resolved and injected per-tool. You could configure it so that an OAuth token gets refreshed and injected into the right parameter automatically, keeping the credential out of the LLM context entirely.

Would be great to see native OAuth proxy support in fastmcp though, especially for the TS side.

ypxing:
Hi @gazzadownunder 
The implementation of #210 doesn't really support what the python version of "OAuth Proxy" has provided - 

```
The OAuth proxy enables FastMCP servers to authenticate with OAuth providers that don’t support Dynamic Client Registration (DCR).
```

It's because the node version just passes through the oauth related data between "MCP client" and "OAuth provider". Two things may not be right -
1. `/oauth/register` is returning `client id/secret` of OAuth provider. It would be **unexpected secret leak** for some people if they are expecting OAuth Proxy to provider its own version of "client" when DCR is not supported by OAuth provider.
2. `invalid_redirect_uri` error - it's due to node version passes the "redirect_uir" of MCP client to OAuth. It doesn't work unless OAuth has supported DCR.

gazzadownunder:
Hi,

The change was based on direct port of the python code/implementation to typescript, which included its limitations. I've noticed the a python version has moved on, give a chance to look at the improvements that have been implemented and how they can be ported.  Once I've had a chance to review the changes, I'll provide an update.

gazzadownunder:
I've created a branch called oauth_proxy-update that ports the latest version of Python fastmcp-oauth and should address your concerns highlighted above.  Please review and test the change to confirm these have been addressed

punkpeye:
Yes — shipped in #210. Built-in OAuth Proxy, same role as the Python one.

[...]

PR #262, opened 2026-04-29
fix(auth): fix OAuthProxy credential leak and redirect_uri validation

gazzadownunder:
Addresses Issue: OAuth Proxy #201 https://github.com/punkpeye/fastmcp/issues/201

[ypxing](https://github.com/ypxing) please test that this addresses your concerns

- Generate proxy-specific client_id/client_secret per DCR registration instead of returning upstream provider credentials to MCP clients
- Add registeredClientsByClientId map for O(1) client lookup
- Fix authorize() and exchangeAuthorizationCode() to validate against proxy-issued client IDs (not the upstream client_id)
- Fix validateRedirectUri() default: undefined → localhost-only (not [])
- Add redirectUris array to ProxyDCRClient for per-client URI validation
- Update all tests to use DCR-returned proxy client_id

Closes CWE-601 open-redirect and upstream credential exposure.

### DIFF
--- a/src/FastMCP.oauth-proxy.test.ts
+++ b/src/FastMCP.oauth-proxy.test.ts
@@ -153,18 +153,19 @@ describe("FastMCP OAuth Proxy Integration", () => {
     });
 
     try {
-      // First register a client
-      await fetch(`http://localhost:${port}/oauth/register`, {
+      // First register a client and capture the proxy-issued client_id
+      const dcrResp = await fetch(`http://localhost:${port}/oauth/register`, {
         body: JSON.stringify({
           redirect_uris: ["https://client.example.com/callback"],
         }),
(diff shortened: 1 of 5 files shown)


ypxing:
Thanks @gazzadownunder 
LGTM ❤️

jabr:
This is pretty major issue... any reason it hasn't been merged yet? 

Without this, my upstream client id is effectively just a public client as well, with the secret shared to anyone that wants it... I can lock down the redirect uri more tightly, but reusing that secret just throws away one layer of security defense for no reason.

edit: That said, I don't think the change to the `allowedRedirectUriPatterns` behavior is necessarily a good idea to include in this MR. I can understand wanting a "better" default (instead of forcing users to set something), but that's a very different issue, imo.

PR #265, opened 2026-05-02
fix(stdio): add stdin close/end listeners to prevent zombie/orphan processes

### LINKED ISSUE #264: stdio transport: server becomes zombie/orphan process when MCP client closes stdin
### Description

When a FastMCP server runs with the `stdio` transport and the MCP client (e.g. Claude Code, any host that spawns the server as a subprocess) closes its end of the stdin pipe — due to a restart, crash, or normal shutdown — **the FastMCP server process keeps running indefinitely** as an orphan/zombie.

In production environments this leads to unbounded accumulation of leaked server processes. The upstream SDK PR (#2003) documents a real-world case of **37 orphaned processes consuming 26,000+ CPU-seconds**.

### Root cause

`StdioServerTransport.start()` in `@modelcontextprotocol/sdk` only registers `data` and `error` listeners on `process.stdin`. It does **not** listen for the `close` or `end` events that Node.js fires when the readable side of the pipe is closed by the remote end. So when the client disappears, the server never learns about it and never calls `transport.close()`.

The same issue is being fixed at the SDK layer in:
- **modelcontextprotocol/typescript-sdk PR #2003** — https://github.com/modelcontextprotocol/typescript-sdk/pull/2003

And has been independently reported downstream:
- **google-tools-mcp issue #26** — https://github.com/karthikcsq/google-tools-mcp/issues/26

### Why FastMCP should also fix this

1. **SDK fix is not yet released** — users on current npm releases are unprotected.
2. **Belt-and-suspenders** — even after the SDK ships the fix, having FastMCP detect the condition independently means the server is resilient to any future regression at the SDK layer.
3. **FastMCP already wraps the transport** — it is the natural place to add this guard.

### Proposed fix

After `await session.connect(transport)` in the `stdio` branch of `FastMCP.start()`, add:

```typescript
const onStdinClose = () => { transport.close().catch(() => {}); };
process.stdin.on("close", onStdinClose);
process.stdin.on("end", onStdinClose);
```

This ensures that when the MCP client closes its end of the pipe, `transport.close()` is called and the server exits cleanly — regardless of which SDK version is installed.

### Steps to reproduce

1. Start any FastMCP stdio server.
2. Spawn it as a child process from a Node script.
3. Call `child.stdin.destroy()` (simulating a client disconnect).
4. Observe: the server process does **not** exit.

### Environment

- FastMCP: latest
- `@modelcontextprotocol/sdk`: any version prior to the fix in PR #2003
- Node.js: any

### References

- SDK fix PR: https://github.com/modelcontextprotocol/typescript-sdk/pull/2003
- Downstream report: https://github.com/karthikcsq/google-tools-mcp/issues/26

ElliotDrel:
## Summary

Fixes #264.

When an MCP client closes its end of the stdin pipe (restart, crash, or normal shutdown), `StdioServerTransport` never fires a close event because the underlying `@modelcontextprotocol/sdk` only registers `data` and `error` listeners on `process.stdin` — not `close` or `end`. The FastMCP server process therefore lingers indefinitely as a zombie/orphan.

This PR adds belt-and-suspenders `process.stdin` listeners immediately after `session.connect(transport)` for the `stdio` transport path. When either `close` or `end` fires, `transport.close()` is called and the process exits cleanly.

## What changed

**`src/FastMCP.ts`** — inside the `if (config.transportType === "stdio")` block, after `await session.connect(transport)`:

```typescript
const onStdinClose = () => { transport.close().catch(() => {}); };
process.stdin.on("close", onStdinClose);
process.stdin.on("end", onStdinClose);
```

## Why here and not just rely on the SDK fix?

- The SDK fix (modelcontextprotocol/typescript-sdk#2003) is **not yet released** — users on current npm versions are unprotected.
- Belt-and-suspenders: even after the SDK ships the patch, having FastMCP detect the condition independently protects against future regressions.
- FastMCP is the natural place for this guard since it owns the `transport` reference and the session lifecycle.

## References

- Closes #264
- SDK upstream fix: https://github.com/modelcontextprotocol/typescript-sdk/pull/2003
- Downstream report: https://github.com/karthikcsq/google-tools-mcp/issues/26

## Test plan

- [ ] Start a FastMCP stdio server as a child process
- [ ] Destroy the child's stdin (`child.stdin.destroy()`) to simulate a client disconnect
- [ ] Confirm the server process exits (previously it would hang)
- [ ] Confirm existing tests still pass (`npm test`)


ElliotDrel:
Hi @punkpeye — just checking in on this. The PR has been CI-green and merge-ready since May 3 (15 days). It's a minimal 3-line fix that adds `close` and `end` listeners to stdin alongside the existing `data`/`error` listeners, preventing zombie processes when an MCP client exits. All review threads resolved. Happy to address any questions!

dlecan:
@punkpeye it would be great to fix that issue quickly
Thanks

PR #266, opened 2026-05-04
feat: add cors option to httpStream config

netanelavr:
mcp-proxy accepts a `cors` parameter but FastMCP does not pass it through.
Custom headers fail CORS preflight and there is no way to fix it from
userland since mcp-proxy handles OPTIONS before Hono middleware runs.

Adds `cors?: boolean | CorsOptions` to the httpStream start options.

### What changed

- `src/FastMCP.ts` - parse and forward `cors` to both `startHTTPServer`
  call sites, re-export `CorsOptions`
- `src/FastMCP.test.ts` - custom allowed headers test, disabled CORS test
- `README.md` - usage example and option reference

### Usage

```ts
server.start({
  transportType: "httpStream",
  httpStream: {
    port: 8080,
    cors: {
      origin: "http://localhost:3000",
      allowedHeaders: ["Content-Type", "X-My-Header"],
      credentials: true,
    },
  },
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,46 @@ openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -node
 
 See the [https-server example](src/examples/https-server.ts) for a complete demonstration.
 
+##### CORS Configuration
+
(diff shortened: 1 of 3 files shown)


punkpeye:
Thanks!

PR #272, opened 2026-05-31
feat: support returning structuredContent from tools

### LINKED ISSUE #81: Add support for structuredContent
Hi, the idea is to support https://github.com/modelcontextprotocol/typescript-sdk/pull/454 to be able to define a tool outputSchema and return structured content.
Thanks

jgarciaweb:
## What

Allows tools to return `structuredContent` alongside `content`, so a tool that declares an `outputSchema` can actually return matching structured output.

## Why

FastMCP already advertises a tool's `outputSchema` in the `tools/list` response, but the result validator `ContentResultZodSchema` is `.strict()` and silently dropped/rejected any `structuredContent` field. The net effect today:

- A tool can declare `outputSchema` but has no supported way to return structured output on success (the only place `structuredContent` ever reached the wire was the `UserError.extras` error path).
- Spec-compliant clients validate a result's `structuredContent` against the declared `outputSchema` (see modelcontextprotocol/typescript-sdk#454, shipped in `@modelcontextprotocol/sdk` 1.x) — so the declared schema had nothing to validate against.

Closes #81.

## Change

- Add an optional `structuredContent?: { [key: string]: unknown }` field to the `ContentResult` type and to `ContentResultZodSchema`.
- `content` stays required, matching the MCP `CallToolResult` shape and the spec's backward-compat guidance (a tool returning structured content should also return a serialized `content` block).

This is a minimal, additive change — the `execute` return union already includes `ContentResult`, so `return { content, structuredContent }` now flows straight through instead of being stripped by the strict schema.

## Usage

```ts
server.addTool({
  name: "get_weather",
  parameters: z.object({ city: z.string() }),
  outputSchema: z.object({ humidity: z.number(), temperature: z.number() }),
  execute: async () => {
    const data = { humidity: 65, temperature: 72 };
    return {
      content: [{ type: "text", text: JSON.stringify(data) }],
      structuredContent: data,
    };
  },
});
```

## Tests

Adds a round-trip test that returns `structuredContent` and asserts it on the `callTool` result — which also exercises the SDK client's output-schema validation. Full `src/FastMCP.test.ts` suite (79 tests) passes; `tsc --noEmit`, `eslint`, and `prettier --check` are clean.

## Docs

Adds a "Returning structured content" section to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -940,6 +940,39 @@ server.addTool({
 });
 ```
 
+#### Returning structured content
+
+Declare an `outputSchema` and return `structuredContent` to give clients a typed
+object in addition to the human-readable `content`. MCP clients surface
+`structuredContent` to the model directly, so it does not have to parse data out of
+a text block. For backward compatibility you should still return a `content` block
+(typically the serialized JSON), as the MCP specification recommends.
+
+```js
+server.addTool({
+  name: "get_weather",
+  description: "Get the weather for a city",
+  parameters: z.object({
+    city: z.string(),
+  }),
+  outputSchema: z.object({
+    humidity: z.number(),
+    temperature: z.number(),
+  }),
+  execute: async (args) => {
+    const data = { humidity: 65, temperature: 72 };
+
+    return {
+      content: [{ type: "text", text: JSON.stringify(data) }],
+      structuredContent: data,
+    };
+  },
+});
+```
+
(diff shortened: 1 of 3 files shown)


punkpeye:
Looks like this was already merged in another PR @jgarciaweb ?

PR #278, opened 2026-06-12
feat(auth): support extra authorization params in OAuthProxy

### LINKED ISSUE #276: OAuthProxy: support extra authorization parameters (e.g. Google's access_type=offline)
## Problem

When using `OAuthProxy` with Google as the upstream provider, Google only issues a `refresh_token` if the authorization request includes `access_type=offline`. Without it, the stored Google access token expires after 1 hour and can never be renewed — every API call fails until the user manually re-authenticates.

`OAuthProxy.redirectToUpstream()` builds the authorization URL with `client_id`, `redirect_uri`, `response_type`, `state`, `scope`, and PKCE — but there's no way to pass provider-specific parameters like `access_type=offline` or `prompt=consent`.

## Current workaround

We monkey-patch `redirectToUpstream` to rewrite the `Location` header:

```typescript
const proxy = oauthProxy as any;
const original = proxy.redirectToUpstream.bind(proxy);
proxy.redirectToUpstream = (transaction: unknown) => {
  const response = original(transaction);
  const location = response.headers.get('Location');
  const url = new URL(location);
  url.searchParams.set('access_type', 'offline');
  url.searchParams.set('prompt', 'consent');
  const headers = new Headers(response.headers);
  headers.set('Location', url.toString());
  return new Response(null, { status: response.status, headers });
};
```

This works but is fragile — it depends on `redirectToUpstream` being synchronous and returning a `Response` with a `Location` header.

## Proposed solution

Add an `extraAuthorizationParams` option to the `OAuthProxy` config:

```typescript
new OAuthProxy({
  upstreamAuthorizationEndpoint: 'https://accounts.google.com/o/oauth2/v2/auth',
  upstreamTokenEndpoint: 'https://oauth2.googleapis.com/token',
  upstreamClientId: '...',
  upstreamClientSecret: '...',
  // ...
  extraAuthorizationParams: {
    access_type: 'offline',
    prompt: 'consent',
  },
});
```

Then in `redirectToUpstream()`, append them to the URL:

```typescript
if (this.config.extraAuthorizationParams) {
  for (const [key, value] of Object.entries(this.config.extraAuthorizationParams)) {
    authUrl.searchParams.set(key, value);
  }
}
```

## Impact

This affects any FastMCP deployment using Google (or other providers that require extra authorization params for refresh tokens) in remote/httpStream mode. Without it, connections die after the access token's TTL (1 hour for Google) and users must manually re-authenticate.

## Environment

- fastmcp: 4.0.2 (also affects 4.0.1)
- Transport: httpStream (remote mode)
- Upstream provider: Google OAuth 2.0
- Platform: Google Cloud Run (serverless)

anneheartrecord:
Closes #276

## Problem

`OAuthProxy.redirectToUpstream()` builds the upstream authorization URL with only the core OAuth parameters. Providers like Google require extra parameters on the authorization request — `access_type=offline` to issue a `refresh_token` at all, and `prompt=consent` to re-issue it on re-auth. Without them, the stored Google access token expires after 1 hour and can never be renewed.

## Solution

Adds an `extraAuthorizationParams` option to `OAuthProxyConfig`, exactly as proposed in the issue:

```typescript
new OAuthProxy({
  // ...
  extraAuthorizationParams: {
    access_type: "offline",
    prompt: "consent",
  },
});
```

### Safety

- Extra params are applied **before** the proxy sets its core parameters, so the proxy-controlled values always win on conflict.
- Reserved keys (`client_id`, `redirect_uri`, `response_type`, `state`, `scope`, `code_challenge`, `code_challenge_method`) are additionally filtered out, so configuration can never override the security-critical core of the request.
- When the option is not set, behavior is unchanged (covered by a test asserting the exact parameter set).

## Testing

- New `src/auth/OAuthProxy.extra-authorization-params.test.ts` (4 tests): params appended; unchanged behavior without config; reserved keys cannot be overridden; proxy PKCE stays intact.
- Full auth suite green (178 tests), `prettier` / `eslint` / `tsc --noEmit` clean.
- Docs updated in `docs/oauth-proxy-guide.md`.

### DIFF
--- a/docs/oauth-proxy-guide.md
+++ b/docs/oauth-proxy-guide.md
@@ -265,6 +265,7 @@ interface OAuthProxyConfig {
   consentRequired?: boolean; // default: true
   consentSigningKey?: string; // auto-generated if not provided
   allowedRedirectUriPatterns?: string[];
+  extraAuthorizationParams?: Record<string, string>; // provider-specific params
   transactionTtl?: number; // seconds, default: 600
   authorizationCodeTtl?: number; // seconds, default: 300
 
@@ -279,6 +280,28 @@ interface OAuthProxyConfig {
   tokenVerifier?: TokenVerifier; // custom JWT verification
 }
 ```
+
+### Extra Authorization Parameters
+
+Some providers require non-standard parameters on the authorization request.
+Google, for example, only issues a `refresh_token` when the request includes
+`access_type=offline` — without it, access expires after one hour and can
+never be renewed:
+
+```typescript
+const authProxy = new OAuthProxy({
+  // ... other config
+  extraAuthorizationParams: {
+    access_type: "offline", // Google: issue a refresh_token
+    prompt: "consent", // Google: re-issue refresh_token on re-auth
+  },
+});
+```
+
+These parameters are appended to the upstream authorization URL. Core OAuth
+parameters managed by the proxy (`client_id`, `redirect_uri`, `response_type`,
+`state`, `scope`, `code_challenge`, `code_challenge_method`) cannot be
+overridden — entries with those keys are ignored.
 
 ### Redirect URI Patterns
 
--- a/src/auth/OAuthProxy.extra-authorization-params.test.ts
+++ b/src/auth/OAuthProxy.extra-authorization-params.test.ts
@@ -1,0 +1,140 @@
+/**
+ * Tests for `extraAuthorizationParams` (issue #276).
+ *
+ * Providers like Google only issue a refresh_token when the authorization
+ * request carries provider-specific parameters (`access_type=offline`,
+ * `prompt=consent`). `OAuthProxy` must be able to append such parameters to
+ * the upstream authorization URL, while never letting them override the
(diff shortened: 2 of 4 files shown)


anneheartrecord:
For reviewers — quick summary to ease review:

**What & why**: Google OAuth needs `access_type=offline` + `prompt=consent` on the authorization request to ever issue a refresh token; without it the access token dies after 1h with no renewal path. `OAuthProxy.redirectToUpstream()` had no way to pass these through.

**Change**: one new optional `extraAuthorizationParams` field on `OAuthProxyConfig`, applied to the upstream auth URL.

**Safety** (the part worth scrutinizing): extras are written *before* the proxy's own core params, and reserved keys (`client_id`, `redirect_uri`, `response_type`, `state`, `scope`, `code_challenge`, `code_challenge_method`) are filtered out — so config can never override the security-critical core or the proxy's PKCE. No behavior change when unset (test-locked).

4 new tests + full auth suite (178) green; prettier/eslint/tsc clean. Happy to adjust naming or the reserved-key policy if you'd prefer a different shape.

PR #286, opened 2026-07-04
fix(auth): persist OAuth proxy state

### LINKED ISSUE #277: OAuthProxy: DCR client registrations and in-flight transactions are not persisted via TokenStorage
## Summary

`OAuthProxy` accepts a pluggable `TokenStorage` (and wraps it in `EncryptedTokenStorage`), but only the JWT-swap state goes through it (`mapping:<jti>`, `upstream:<key>`). Two other classes of OAuth state live in plain in-memory `Map`s on the proxy instance:

- **DCR client registrations** — `registeredClients` / `registeredClientsByClientId`, written by `registerClient()`
- **In-flight OAuth transactions** — `transactions`, written between `authorize()` and the callback

Observed in 4.0.2 (`dist/chunk-*.js`; the maps are initialized on the class and never touch `this.tokenStorage`).

## Impact

For any deployment that restarts (deploys) or runs more than one instance:

1. **Every restart invalidates every registered client.** A client that registered via DCR (e.g. claude.ai's connector flow) gets `invalid_client` on its next `/oauth/authorize` or `/oauth/token` call after the server restarts, because the registration only existed in process memory. Users have to re-register/reconnect after every deploy.
2. **Multiple replicas break flows even with pinned keys + shared TokenStorage.** With `jwtSigningKey`/`encryptionKey` pinned and a network-backed `TokenStorage`, issued tokens verify across instances — but a client registered on instance A is unknown to instance B, and an authorization flow started on A fails when the callback lands on B (transaction not found).

## What we'd like

Move DCR registrations and transactions behind the same `TokenStorage` seam the token mappings use (e.g. `client:<id>` / `transaction:<state>` keys), or expose a documented persistence hook for them. That would make a stateless multi-replica deployment fully workable with a shared store.

## Current workaround

- Wrapping `registerClient()` to write each registration through to our own store, and re-hydrating the proxy's maps from it at boot — covers restarts.
- Transactions can't be reached from outside the proxy, so we run a single replica until this is addressed.

Happy to contribute a PR if moving this state into `TokenStorage` is a direction you'd accept.

JSap0914:
## Summary

- Persist OAuthProxy DCR clients, in-flight transactions, and generated authorization codes through the configured TokenStorage
- Keep the existing in-memory maps as local caches while allowing another proxy instance to hydrate the state from storage
- Add regression coverage for DCR authorization and authorize -> callback -> token exchange across separate instances sharing encrypted TokenStorage

Fixes #277.

## Validation

- pnpm vitest run src/auth/OAuthProxy.storage-persistence.test.ts
- pnpm vitest run src/auth/OAuthProxy.open-redirect.test.ts
- pnpm vitest run src/auth/OAuthProxy.token-swap.test.ts
- pnpm vitest run src/auth/OAuthProxy.extra-authorization-params.test.ts src/auth/OAuthProxy.token-parsing.test.ts
- pnpm exec prettier --check .
- pnpm exec eslint .
- pnpm exec tsc --noEmit
- pnpm exec jsr publish --dry-run --allow-dirty
- pnpm lint

Note: the first full `pnpm test` run passed all 369 tests but failed on unrelated unhandled HTTP transport errors. A second full run reproduced unrelated `FastMCP.completions.test.ts` timeouts under the parallel suite; that file passed when run in isolation.

### DIFF
--- a/docs/oauth-advanced-features.md
+++ b/docs/oauth-advanced-features.md
@@ -311,9 +311,10 @@ const encryptionKey = await JWTIssuer.deriveKey(
 
 ### Storage Backend Selection
 
-- **MemoryTokenStorage**: Fast but loses data on restart. Good for development.
-- **DiskStore**: Persistent but slower. Good for single-server production.
+- **MemoryTokenStorage**: Fast but loses data on restart, and per-process. Good for development.
+- **DiskStore**: Persistent but slower. Good for single-host production, including several processes sharing one directory.
 - **EncryptedTokenStorage**: Additional overhead for encryption. Use when storing sensitive data.
+- **Custom (Redis/database)**: Required to run instances on more than one host. Must implement `take` — see [Running Multiple Instances](oauth-proxy-guide.md#running-multiple-instances).
 
 ### Cleanup Intervals
 
--- a/docs/oauth-proxy-features.md
+++ b/docs/oauth-proxy-features.md
@@ -203,8 +203,10 @@ Ready-to-use provider implementations:
 
 Background processes maintain system health:
 
-- Expired transactions automatically removed
-- Authorization codes deleted after use or expiration
+- Transactions and authorization codes are written with a TTL derived from their
+  expiry, and reclaimed by the storage backend
+- Authorization codes are consumed on first use; a small marker is kept until
+  the code would have expired, so a replay reports "already used"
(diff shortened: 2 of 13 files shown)


punkpeye:
Documentation needs updating

PR #292, opened 2026-07-11
Pass load context (client info, log, session) to resource and prompt load()

### LINKED ISSUE #171: Add a way to get the client name and version within a tool execution
We have a use case for this, could it be added into the second context param that's passed to `execute()`?

Starlight143:
## Summary

Issue #171 was resolved for **tools** by #173, which added `client.version` / `getClientVersion()` to the `Context<T>` object passed to `tool.execute`. The issue was left open because a follow-up comment (attiks, 2025-09-12) pointed out that `Resource.load`, `ResourceTemplate.load`, and `Prompt.load` never receive a `Context`, so resources and prompts have no way to read client name/version, log through the MCP logging channel, or access session / session ID / request ID.

I re-checked `src/FastMCP.ts` on current `main` (b170e78) before starting: `InputPrompt.load` / `Prompt.load` were `(args, auth?) => ...`, `InputResourceTemplate.load` / `ResourceTemplate.load` were `(args, auth?) => ...`, and `Resource.load` was `(auth?) => ...` — none received anything like `Context`. This PR closes that gap.

## What changed

- Added `LoadContext<T>`, defined as `Omit<Context<T>, "reportProgress" | "streamContent">`, and exported it alongside `Context`.
  - `reportProgress` and `streamContent` are intentionally excluded: both are tied to a tool call's progress token / streaming notification, which resource and prompt reads do not have.
- `Resource.load`, `ResourceTemplate.load`, and `Prompt.load` (and their `Input*` counterparts) now accept an additional, optional trailing `context?: LoadContext<T>` parameter:
  - `Resource.load: (auth?: T, context?: LoadContext<T>) => ...`
  - `ResourceTemplate.load` / `Prompt.load`: `(args, auth?: T, context?: LoadContext<T>) => ...`
- `FastMCPSession` gained two private helpers, `#createLoadContext()` and `#createLog()`, so the `client`/`log`/`session`/`sessionId`/`requestId` construction is shared between the tool, resource, and prompt handlers instead of being duplicated (the tool handler's inline `log` object was refactored to call `#createLog()`).
- Updated the three call sites (`GetPromptRequestSchema`, and the two `resource.load` / `resourceTemplate.load` calls inside `ReadResourceRequestSchema`) to pass the new context through.
- README: added `auth`/`context` notes and short examples for Resources, Resource templates, and Prompts.

## Backward compatibility

The new parameter is additive and appended **after** the existing `auth` parameter, so:
- Existing `load` implementations with fewer parameters (`load()`, `load(auth)`, `load(args)`, `load(args, auth)`) keep compiling and behaving exactly as before — JS/TS callers are free to ignore extra arguments.
- No existing public API, type export, or runtime behavior for tools was touched (the tool `Context` shape is unchanged; only the `log` object construction was extracted into a shared private method).
- `FastMCP.embedded()` (the top-level convenience method, not a request handler) still calls `.load()` with no context, since it has no active request/session to source one from — this is unchanged.

## Testing

- Added 3 new tests: resources / resource templates / prompts reading `context.client.version`, plus assertions that `reportProgress` and `streamContent` are *not* present on the load context (locking in the intentional subset).
- Updated 4 pre-existing "provides auth to ..." tests whose `vi.fn()` mocks assert `toHaveBeenCalledWith(...)` — added `expect.anything()` for the new third argument.
- `npm run build` (tsup): passes.
- `npx vitest run`: 370/370 tests pass.
- `npx eslint .` and `npx prettier --check` on the touched files: clean.
- `npx tsc --noEmit`: clean.

(Note: I used `npm` instead of `pnpm` to install dependencies locally — `pnpm install` crashed with a native `STATUS_STACK_BUFFER_OVERRUN` on this machine, unrelated to this change. `pnpm-lock.yaml` was not touched.)

Closes #171.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1329,6 +1329,23 @@ async load() {
 }
 ```
 
+`load` also receives `auth` (the value returned by your `authenticate` function, if any) and a `context` object as its second and third arguments. `context` mirrors the `client`, `log`, `session`, and `sessionId` fields available to `tool.execute` (see [Session ID and Request ID Tracking](#session-id-and-request-id-tracking)); `reportProgress` and `streamContent` are not included since they are tied to a tool call's progress token:
+
+```ts
+server.addResource({
+  uri: "file:///logs/app.log",
+  name: "Application Logs",
+  mimeType: "text/plain",
+  async load(auth, context) {
+    context.log.info("loading application logs", { requestedBy: auth?.userId });
+
+    return {
+      text: await readLogFile(),
+    };
+  },
+});
+```
+
 ### Resource templates
 
 You can also define resource templates:
@@ -1352,6 +1369,8 @@ server.addResourceTemplate({
   },
 });
 ```
+
+Like plain resources, `load` also receives `auth` and `context` as its second and third arguments (see [Resources](#resources)).
 
 #### Resource template argument auto-completion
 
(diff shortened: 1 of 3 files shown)


punkpeye:
Thank you

PR #328, opened 2026-08-15
feat(core): pass AbortSignal in tool execution context for cancellation and timeouts

hktitof:
### Summary
Adds first-class `AbortSignal` support to the tool execution `Context`.

### Motivation
Long-running tools (e.g. database queries, external API calls, sub-processes) currently have no mechanism to observe client cancellations or execution timeouts. When a client closes an SSE stream or a tool exceeds `timeoutMs`, background tasks continue executing as zombie operations.

### Changes
- Added `signal: AbortSignal` to `Context` passed into `tool.execute(args, context)`.
- Forwards client cancellation/disconnects (`extra.signal`) to the tool's `signal`.
- Triggers `signal.abort()` automatically when `tool.timeoutMs` triggers.
- Guaranteed cleanup of event listeners on promise settlement in `.finally()`.
- Backward compatible with existing tools and `LoadContext`.
- Added test coverage in `src/toolCancellation.test.ts`.

### Verification
- `npx vitest run src/toolCancellation.test.ts src/FastMCP.test.ts`: 108/108 passed.
- `tsc --noEmit`: Clean (0 errors).
- Zero formatting churn across untouched lines.


punkpeye:
Rebased onto `main` and pushed a follow-up commit. Thanks for this — the gap was real.

What changed:

- **Disconnects didn't actually work.** The SDK only aborts `extra.signal` for an explicit `notifications/cancelled`; `Protocol._onclose()` and `StreamableHTTPServerTransport` never touch it. Added a session-scoped controller aborted from `close()` and `server.onclose`, so a dropped transport reaches the tool.
- **Replaced the manual forwarding with `AbortSignal.any()`.** No listener bookkeeping, and nothing leaks when `execute` throws synchronously (the old `.finally()` never attached in that case).
- **Timeout now aborts with the same `UserError` the caller gets**, so tools can tell a timeout from a cancellation. Kept the clearable timer so a completed call never aborts retroactively — cleanup attached to `signal` stays safe.
- **Tests**: added client-cancellation, transport-drop, sync-throw, and no-retroactive-abort cases. Renamed to `FastMCP.tool-cancellation.test.ts` to match the other files.
- **README**: documented `context.signal`, including that an HTTP client vanishing mid-request still isn't detectable — use `timeoutMs` for that.

`pnpm test` 568/568 · `pnpm lint` clean.

PR #330, opened 2026-08-18
fix: cancel media response bodies on HTTP errors

pacocartones:
When imageContent/audioContent receive a non-2xx response, they throw
before consuming or cancelling the response body. This leaves an unread
ReadableStream owned by the fetch implementation.

Cancel the body before preserving the existing HTTP error. Cancellation
failures are intentionally ignored so callers still receive the original
status/statusText diagnostic.

Tests cover:

- image URL timeout and default timeout message
- audio URL timeoutMs override
- HTTP error propagation and body cancellation for image/audio

Full FastMCP test suite passes.


punkpeye:
Verified the leak is real — 20 sequential fetches against a 503-with-body endpoint left 20 sockets open on Node 24; zero with `cancel()`. Good catch.

Pushed cleanups to your branch:

- Extracted the helper to `src/cancelResponseBody.ts` and applied it to `DiscoveryDocumentCache` too — it had the identical leak at the `!res.ok` throw.
- Dropped the two timeout tests; `FastMCP.media-timeout.test.ts` already covered the default message, the AbortSignal, and `timeoutMs`.
- Moved the HTTP-error test out of the 6k-line `FastMCP.test.ts` into that file, renamed `FastMCP.media-fetch.test.ts` now that it covers more than timeouts.

Merging once CI is green.

PR #331, opened 2026-08-20
fix(edge): reject a replayed stream that collides with one already live

pacocartones:
## What's wrong

`WebStreamableHTTPServerTransport.handleReplayEvents` (the resumability path for a GET
reconnect carrying `Last-Event-Id`) registers the new stream's writer in `_streamMapping`
without checking whether a connection for that same `streamId` is already live. If a second
reconnect resolves to a `streamId` that's still tracked, it silently overwrites the first
writer instead of rejecting the conflict:

```ts
const first = await transport.handleRequest(getRequestWithLastEventId);   // 200, writer A
const second = await transport.handleRequest(getRequestWithLastEventId);  // 200, writer B — should be 409
```

The first connection's stream is now orphaned: no more bytes, no close, no signal to the
client that anything went wrong.

## Root cause

The equivalent guard already exists for the standalone stream (line ~332) and in the official
SDK's Node transport (`@modelcontextprotocol/sdk`, `streamableHttp.js`, `replayEvents`), which
checks `this._streamMapping.get(streamId) !== undefined` via `EventStore.getStreamIdForEventId`
before touching the mapping. `handleReplayEvents` in this file never calls
`getStreamIdForEventId` at all — the conflict check that exists one method up simply isn't
mirrored here.

## The fix

Before creating the replay stream, if `EventStore.getStreamIdForEventId` is implemented (it's
optional on the interface), resolve the `streamId` the reconnect would land on and, if it's
already tracked in `_streamMapping`, answer `409` — same status and error shape as the
standalone-stream guard. If the store doesn't implement `getStreamIdForEventId`, this degrades
to the previous (racy) behavior, matching what the SDK's own transport does in that case.

## Tests

New test in `src/edge/WebStreamableHTTPServerTransport.test.ts`: two `GET` reconnects with the
same `Last-Event-Id` against a store whose `getStreamIdForEventId` always resolves to the same
`streamId`. First reconnect: `200`, writer tracked. Second: `409`, and the tracked writer is
still the first one (no silent swap).

`npx vitest run src/edge/`: 42/42 passing (2 files). No other suites touched.

### DIFF
--- a/src/edge/WebStreamableHTTPServerTransport.test.ts
+++ b/src/edge/WebStreamableHTTPServerTransport.test.ts
@@ -264,6 +264,37 @@ describe("WebStreamableHTTPServerTransport", () => {
     await first.body?.cancel("test cleanup");
   });
 
+  it("should reject a colliding replay from a store without getStreamIdForEventId", async () => {
+    // `getStreamIdForEventId` is optional and most stores do not implement it,
+    // so a guard that only runs when it exists would leave the majority of
+    // stores overwriting live writers exactly as before. The id
+    // `replayEventsAfter` returns is the one the writer is registered under,
+    // and checking that one needs no cooperation from the store at all.
+    const transport = new WebStreamableHTTPServerTransport({
+      enableJsonResponse: true,
+      eventStore: {
+        replayEventsAfter: async () => "resumed-stream",
+        storeEvent: async () => "evt-1",
+      },
+      sessionIdGenerator: () => "test-session",
+    });
+    await createServer().connect(transport);
+    await transport.handleRequest(createInitializeRequest("application/json"));
+
+    const internals = transport as unknown as TransportInternals;
+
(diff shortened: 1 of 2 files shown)


punkpeye:
Good catch — merging with one addition.

Your pre-check only runs for stores implementing the optional `getStreamIdForEventId`, and most don't, so those would still overwrite silently. Added a second check against the id `replayEventsAfter` actually returns (the one `trackStream` registers under), with no await between it and the `set`. Covers stores without the lookup, stores whose two methods disagree, and a reconnect racing past the pre-check.

Also aligned the 409 message with the standalone guard, and noted why we skip the SDK's `400 Invalid event ID format` branch.

Unrelated to this PR, but found while reviewing: `handleReplayEvents` deadlocks whenever the store replays *any* event — a TransformStream's readable has a high-water mark of 0, and nothing reads it until the Response is returned, so the first `writer.write()` parks forever. Every existing test passes only because its store replays zero events. Tracking that separately.

PR #337, opened 2026-08-24
fix: stop pumping a custom route stream once the client is gone

pacocartones:
A route added via `getApp()` may return a body that never ends on its own — an SSE stream, or a proxied upstream. The pump that copies that body into the Node response loops until the source says `done`, which for such a route only happens if the source itself stops.

When the client disconnects, the loop therefore keeps reading forever into a response nobody can receive. The `finally` block releases the reader lock without cancelling the body, so the route is never told to stop producing and keeps holding whatever it holds — an upstream connection, a timer, a subscription.

Stop when the response can no longer be written to, and cancel the body on the way out.

The added test registers a route whose stream never completes, reads one chunk, aborts the request, and asserts the stream's `cancel()` runs. Without the change it never does.

Suite on the branch: 47 files / 593 passed (592 on a clean `main`). Two pre-existing failures were verified to be identical on `main` and are untouched here: `tsc --noEmit` on `src/edge/WebStreamableHTTPServerTransport.ts` (TS2345), and `prettier --check src/FastMCP.ts` (version drift reformats unrelated unions — I deliberately kept only my own hunk rather than reformatting them).

### DIFF
--- a/src/FastMCP.routes.test.ts
+++ b/src/FastMCP.routes.test.ts
@@ -1038,3 +1038,67 @@ test("custom route stream stops when the client disconnects", async () => {
     await server.stop();
   }
 });
+
+test("custom route stream stops when a quiet client disconnects", async () => {
+  const port = await getRandomPort();
+
+  const server = new FastMCP({
+    name: "Test",
+    version: "1.0.0",
+  });
+
+  const app = server.getApp();
+
+  let cancelled = false;
+
+  app.get("/quiet-events", () => {
+    const encoder = new TextEncoder();
+    let sent = false;
+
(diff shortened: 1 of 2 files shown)


punkpeye:
Merged with one addition on top.

The loop only re-checked `res.destroyed` *after* `reader.read()` resolved, so it still needed the source to produce a chunk to notice the disconnect. A real SSE route between heartbeats is quiet at exactly that moment — it never got cancelled.

Now each read races a `close` promise on `res`. Added a companion test (stream emits one chunk, then goes silent forever) that fails on your original commit.

PR #345, opened 2026-08-29
fix: complete all enum values when the prompt argument value is empty

pacocartones:
### Summary

When a prompt argument declares `enum`, FastMCP auto-completes it with a fuzzy `Fuse` index. `Fuse.search("")` returns zero matches, so when a client requests completions with an **empty value** — the standard way editors surface the valid values before the user has typed anything — the response is an empty list, even though the server declared the full valid set.

### Change

In the `addPrompt` completion handler, when the requested value is `""`, return the full `enums[name]` list. The existing central `capCompletionValues` still trims to the MCP 100-item cap and sets `hasMore`, so large enums stay spec-compliant. Non-empty values keep the existing fuzzy path unchanged.

### Test

Adds `"offers all enum values when the completion value is empty"` to `src/FastMCP.test.ts` (RED before the change, GREEN after); the existing non-empty enum test still passes. Full suite green (`pnpm test`), `tsc`/`eslint`/`prettier` clean.

### DIFF
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -6037,3 +6037,43 @@ test("exports SessionError and FastMCPError from the package root", () => {
   // so the base class can serve as a catch-all for all fastmcp errors.
   expect(new UserError("u")).toBeInstanceOf(FastMCPError);
 });
+
+test("offers all enum values when the completion value is empty", async () => {
+  await runWithTestServer({
+    run: async ({ client }) => {
+      const response = await client.complete({
+        argument: {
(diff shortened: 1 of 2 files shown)


punkpeye:
Verified: `Fuse.search("")` returns zero matches, so a client asking for completions before the user types anything got an empty list despite the server declaring the full enum. Your test is red on main, green with the fix.

Returning the full enum is right, and `capCompletionValues` already handles the 100-item cap and `hasMore`. Merging.

PR #347, opened 2026-08-30
feat(oauth): add Client ID Metadata Document (CIMD) support to OAuthProxy

LordNayan:
Adds enableCimd (default: false) to OAuthProxyConfig. When enabled,
authorize() and exchangeAuthorizationCode() resolve a client_id that
is an HTTPS URL by fetching it (SEP-991), instead of requiring a prior
POST /oauth/register — the fetched document is validated (self-referential
client_id, redirect_uris checked against the existing
allowedRedirectUriPatterns allow-list, no-loopback/private-host guard)
and cached through the existing state store exactly like a DCR client.
getAuthorizationServerMetadata() advertises
client_id_metadata_document_supported when enabled.

Fixes MCP clients (VS Code among them) that require CIMD and don't
fall back to DCR. Off by default; no behavior change otherwise.

Adds src/auth/utils/cimd.ts + unit tests, and
src/auth/OAuthProxy.cimd.test.ts integration tests.

### DIFF
--- a/docs/oauth-proxy-guide.md
+++ b/docs/oauth-proxy-guide.md
@@ -267,6 +267,7 @@ interface OAuthProxyConfig {
   consentRequired?: boolean; // default: true
   consentSigningKey?: string; // auto-generated if not provided
   allowedRedirectUriPatterns?: string[];
+  enableCimd?: boolean; // default: false — accept Client ID Metadata Documents
   extraAuthorizationParams?: Record<string, string>; // provider-specific params
   transactionTtl?: number; // seconds, default: 600
   authorizationCodeTtl?: number; // seconds, default: 300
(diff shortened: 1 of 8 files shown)


punkpeye:
Thanks — this is a careful piece of work: self-referential `client_id` check, allow-list reuse, SSRF guard, bounded response read, and thorough tests.

One thing I pushed a fix for: resolved CIMD clients were cached permanently (`saveRegisteredClient` takes no TTL, and the in-memory map never evicts). That breaks the invariant the cache is built on — a DCR registration is immutable, a CIMD document is not. A client rotating a compromised redirect URI could never revoke it here.

My commit:
- stamps a 15-minute `expiresAt` on a resolved client; the state store drops lapsed copies from memory and storage
- routes every client lookup through one `resolveClient()` helper (also replaces the duplicated resolve+cache blocks), so a lapse always means "re-fetch", never "unknown client" mid-flow — the callback check went through `isTransactionCallbackRegistered` and would otherwise have failed a legitimate flow
- adds `clientIdMetadataDocumentSupported` to the `oauth.authorizationServer` type in `FastMCP.ts`
- documents `enableCimd` in `docs/oauth-proxy-guide.md`
- adds two tests (both red without the TTL)

Merging once CI is green.

LordNayan:
Thank you @punkpeye , now I can finally remove the patch I was using in my repo and replace it with the native functionality of FastMCP. Thanks for being super quick with this.

PR #243, opened 2026-02-16
feat: add OpenAPI-to-MCP converter (fromOpenAPI) — closes #42

lucamorettibuilds:
## Summary

Implements #42 — automatic conversion of OpenAPI specs to MCP tools, resources, and resource templates.

### `src/openapi.ts` — OpenAPI spec parser
- Inline OpenAPI 3.x type definitions (zero external dependencies)
- `extractRoutes()` — extract all operations from an OpenAPI spec
- `buildToolParameters()` — convert OpenAPI schemas to JSON Schema for tool params
- `classifyRoute()` — classify routes as `resource`, `resourceTemplate`, or `tool`
- Full `$ref` resolution for schemas, parameters, and request bodies

### `src/fromOpenAPI.ts` — High-level converter
- `fromOpenAPI({ spec, client })` → `{ tools, resources, resourceTemplates }`
- GET without path params → MCP resource
- GET with path params → MCP resource template
- POST/PUT/PATCH/DELETE → MCP tool
- Path parameter substitution, query string building, header extraction
- Request body flattening with `body_` prefix convention
- Pluggable HTTP client interface

### `src/openapi.test.ts` — Tests
- Route extraction, classification, parameter building
- `$ref` resolution
- Request body handling

## Usage

```typescript
import { fromOpenAPI } from "fastmcp/openapi";

const spec = JSON.parse(fs.readFileSync("openapi.json", "utf-8"));
const { tools, resources, resourceTemplates } = fromOpenAPI({
  spec,
  client: {
    request: async (config) => {
      const res = await fetch(config.url, {
        method: config.method,
        headers: config.headers,
        body: config.body ? JSON.stringify(config.body) : undefined,
      });
      return { status: res.status, data: await res.json() };
    },
  },
});

const mcp = new FastMCP({ name: "My API", version: "1.0.0" });
tools.forEach((t) => mcp.addTool(t));
resources.forEach((r) => mcp.addResource(r));
```

## Design decisions
- **Zero new dependencies** — OpenAPI types are inlined to keep the dependency tree clean
- **Pluggable HTTP client** — Users bring their own fetch/axios/etc via a simple interface
- **Flat body params** — Request body properties are prefixed with `body_` to avoid collisions with path/query params
- **Conservative classification** — Only GET routes become resources; everything else is a tool

PR #349, opened 2026-09-01
feat(openapi): add fromOpenAPI() converter v1 for OpenAPI 3.x specs + real-world benchmark suite

### LINKED ISSUE #42: Please implement OpenAPI to MCP converter in the TypeScript version

LordNayan:
## Summary

Adds `fromOpenAPI()` — converts an OpenAPI 3.x document into an MCP
server, one tool per operation:

```ts
import { fromOpenAPI } from "fastmcp/openapi";

  spec: "https://petstore3.swagger.io/api/v3/openapi.json",
  include: (operation) => operation.tags.includes("pet"),
});

await server.start({ transportType: "stdio" });

```

Closes #42. Two prior attempts (#193, #243) were closed as "needs a lot more work," without an itemized list. A [detailed follow-up comment](https://github.com/punkpeye/fastmcp/issues/42) laid out five concrete gaps, based on a production Python converter benchmarked against real specs. This PR addresses each one directly rather than porting the Python `fastmcp`'s own OpenAPI support, which — I confirmed by reading it — has the same four gaps.

### What's included

* External `$ref` resolution — multi-file specs (one YAML per resource, entry doc is pure `$ref`) via `@apidevtools/swagger-parser`, called with the spec's URL/path directly so refs resolve against the right base. Covers relative paths, absolute URLs, and the `other.yaml#/fragment` form.
* Relative `servers[0].url` resolution — e.g. Petstore's own `servers: [{ url: "/api/v3" }]` is resolved against the spec's own origin, not passed through verbatim. Verified against the real, live Petstore v3 API (not just schema validation).
* Explicit tool selection — `include`/`exclude` predicates, a `maxTools` cap, and deprecated-by-default exclusion. Throws instead of silently generating one tool per operation past a 40-operation default:

```ts
await fromOpenAPI({
  spec: "https://api.example.com/openapi.json",
  include: (op) => op.tags.includes("orders"),
  exclude: (op) => op.operationId === "deleteAllOrders",
});

```

* Parameter flattening with a decided collision rule — path, query, header, and body can declare the same name; a colliding non-body param is suffixed `{name}__{location}`, a body property always keeps its bare name.
* A test that can fail — `fromOpenAPI.test.ts` does a real `initialize` + `tools/list` + tool call against the public Petstore v3 API, not just schema validation.
* Auth via `headers` (static or a token-refreshing function), a `baseUrl` override, `mcpNames` for renaming tools, and registering onto an existing `FastMCP` instance via `server`.

### Benchmark suite (src/openapi/fromOpenAPI.benchmark.test.ts)

- Vendors 6 real OpenAPI specs under src/openapi/__fixtures__/benchmark-specs/ (~20MB, refreshable via pnpm bench:openapi:refresh → scripts/refresh-openapi-benchmark-specs.mjs) so the suite is deterministic and network-free.

- For Petstore / Box / Twilio / PostHog / Stripe / the multi-file YAML fixture: converts the entire spec (maxTools: Infinity), then asserts the resulting tool count matches the spec's non-deprecated operation count, every tool name is unique, and every generated schema compiles through the same jsonSchemaAdapter → AJV path FastMCP uses at runtime.

- Slack gets its own case: their public spec is actually Swagger 2.0 (not OpenAPI 3.x, despite the "v2" in the filename referring to their Web API version) — the test pins that fromOpenAPI() correctly rejects it with the documented error, rather than faking a 3.x spec that doesn't exist.

- Runs as its own CI job (openapi-benchmark in .github/workflows/feature.yaml, via pnpm bench:openapi), separate from pnpm test, so large-spec conversion time doesn't slow the main test loop.

### What's left out of v1

Kept deliberately tight so the diff stays reviewable:

* Tools only — every operation becomes a tool, including `GET`s. No resource/resource-template mapping yet (agent clients call tools far more reliably than they read resources, so this was the intentional default, not a gap).
* `application/json` request bodies only — no multipart or form-urlencoded.
* Common query parameter styles only — array params are sent as repeated keys (`?tag=a&tag=b`, OpenAPI's `form`/`explode: true` default). `deepObject`, `spaceDelimited`, `pipeDelimited` aren't implemented.
* OpenAPI 3.x only — Swagger 2.0 is rejected with a clear error.
* No response/`outputSchema` validation — a tool's result is the raw response body (pretty-printed if JSON).
* No CLI — `fromOpenAPI` is a programmatic API; no `fastmcp openapi <spec>` command yet.

Full reference: docs/openapi.md

### Test plan

* `pnpm test` — 34 new tests across ref resolution (incl. a local multi-file YAML fixture), parameter flattening/collisions, naming/collisions, selection/capping, request building (incl. relative-server-URL resolution), and two end-to-end tests (one offline/mocked, one live against Petstore v3)
* `pnpm run lint` (prettier, eslint, `tsc --noEmit`) clean
* Manual smoke test via the MCP in-memory transport against the live Petstore v3 spec — correct tool list, schema, and a real `getPetById` call

### Test Runs

$ pnpm bench:openapi

 RUN  v4.0.15

 ✓ fromOpenAPI benchmark: 'Petstore' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 167ms

 ✓ fromOpenAPI benchmark: 'Box' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 641ms

 ✓ fromOpenAPI benchmark: 'Twilio' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 407ms

 ✓ fromOpenAPI benchmark: 'PostHog' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 11386ms
 
 ✓ fromOpenAPI benchmark: 'Stripe' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 2803ms

 ✓ fromOpenAPI benchmark: 'Multi-file YAML spec' > converts every non-deprecated operation into a uniquely-named tool with a well-formed JSON Schema 118ms

 ✓ fromOpenAPI benchmark: Slack's real spec is Swagger 2.0, which is rejected 27ms

 Test Files  1 passed (1)

Tests  7 passed (7)

$ pnpm test

 Test Files  60 passed (60)

Tests  698 passed (698)

Duration  20.73s


punkpeye:
Reviewed and pushed 4 commits to this branch. Approach and code are solid — this addresses the five gaps from #42 concretely, and the docs are honest about scope. Notes on what I changed:

**1. Bug: schema properties named `nullable` were being dropped.** `normalizeNullable` ran on every object node, so it couldn't tell the OpenAPI keyword from a field with the same name:

```
in:  {type:"object", properties:{nullable:{type:"boolean"}, name:{type:"string"}}}
out: {type:"object", properties:{name:{type:"string"}}}
```

Same class of issue for `$ref`, and for `example`/`default` payloads. The walker is now mode-aware (schema / name-to-schema map / instance data). 3 regression tests added — they fail on the previous commit.

**2. Vendored specs → fetched on demand.** The 20MB was ~3MB permanent growth on every clone. They're now pulled from `benchmark-specs/sources.json` and verified against pinned sha256s, so a run is still reproducible; a spec drifting upstream fails loudly instead of silently changing what's tested. `pnpm test:openapi:refresh` re-pins. `petstore.json` + the multi-file fixture stay committed so part of the suite runs offline. CI caches on the manifest hash.

**3. Live test out of `pnpm test`.** It was the first network test in the main job — a Petstore outage would have failed CI on every unrelated PR. Moved to the `OpenAPI (network)` job alongside the benchmark.

It's also hardened: `getPetById(10)` asserts pet 10's contents on a publicly-writable demo, so anyone deleting that pet breaks the suite. It now accepts an API-level error as still proving the point (a response from the right host means the base URL resolved) and only fails on a URL/fetch error, which is the actual regression it guards. Worth knowing: `findPetsByStatus` is returning 500 upstream right now — I tried it as the stable alternative first.

**4.** Dropped `.claude/settings.local.json` from the diff.

Both jobs green. Diff is now 2,721 additions / 32 files, down from 562,922 / 35.

One thing I want your read on before merging: this ships **tools-only**, while #42 asked for the Python mapping (GET → resource, GET+path params → resource template). Your reasoning in the description is sound and I lean the same way, but it's a public API shape that's awkward to change later. Fine to keep as-is — just confirming it's deliberate rather than deferred.

Also note the PR description still references `pnpm bench:openapi` and vendored specs — worth a quick update.

LordNayan:
It was actually deferred for v2. I thought the PR was too big so we can do it in the next version. Would you want me to implement this in the same version?

punkpeye:
Merging this as is. Thank you. A follow up is welcome.

PR #350, opened 2026-09-01
feat(openapi): fix form-urlencoded request bodies, add GET-to-resource mapping

LordNayan:
## Summary

Two changes to `fromOpenAPI()`, both found by actually running the
benchmark suite's real specs through it rather than just schema-validating
the output:

**1. Form-urlencoded request bodies (bug fix)**

`extractBodyProperties` only ever read `content["application/json"]`. I
checked the actual specs in our own benchmark suite against that:

| Spec   | Ops with a body | `application/json` | form-urlencoded only |
|--------|-----------------:|--------------------:|----------------------:|
| Stripe | 588              | 0                    | 587                    |
| Twilio | 62               | 0                    | 62                     |
| Box    | 118              | 104                  | 3 (+10 other unsupported) |

Every Stripe and Twilio write operation was silently generated as a tool
with **zero body parameters** — no error, no warning, just an unusable
tool. The benchmark suite's assertions (tool count, unique names, "AJV
validates `{}`") didn't catch it, because an empty `properties: {}` schema
trivially passes all three.

Now:
- `application/json` and `application/x-www-form-urlencoded` bodies are
  both flattened (JSON preferred if a route declares both).
- A body declared only in a content type this module still can't handle
  (`multipart/form-data`, `application/json-patch+json`,
  `application/octet-stream`, ...) is **skipped**, not silently turned into
  a broken tool — `fromOpenAPI` logs one `console.warn` listing everything
  it skipped.
- The benchmark suite got a real regression assertion: for any non-GET tool
  whose route declares a body with real properties, the tool's schema must
  have *more* properties than the route's own parameter list. This is the
  check that should have caught the bug — it fails against the old code for
  every Stripe/Twilio operation and passes now.

**2. `resources: true` — GET → resource / resource template (new, opt-in)**

```ts
  spec: "https://petstore3.swagger.io/api/v3/openapi.json",
  include: (op) => op.tags.includes("pet"),
  resources: true,
});
// getPetById (GET /pet/{petId}) → resource template at
// openapi://getPetById/pet/{petId}, readable via client.readResource()
```

An eligible GET becomes a static resource (no parameters) or a resource template (path/query parameters), reusing the existing parameter-flattening and HTTP-execution logic unchanged — only the MCP primitive it's registered as differs. A GET stays a tool when it has any header/cookie parameter (can't be expressed in a resource URI) or any array-typed path/query parameter (OpenAPI's and RFC 6570's array representations don't match). No per-operation override; default is unchanged (resources defaults to false).

### Known limitations (unchanged from v1 except where noted)

* Form-urlencoded nested objects/arrays are JSON-stringified into a single value rather than expanded with bracket notation (Stripe's metadata[key]=value style) — correct for the flat scalar properties that make up the large majority of real form-encoded APIs.
* deepObject/spaceDelimited/pipeDelimited query styles, Swagger 2.0, response/outputSchema validation, and a CLI subcommand remain deferred.

### Test plan

* pnpm test — 40+ new/updated unit tests (form-urlencoded flattening, unsupported-content-type detection, form-body serialization, resource/resource-template eligibility and URI building)
* pnpm test:openapi — hardened benchmark suite passes against real Petstore/Box/Twilio/PostHog/Stripe specs + Slack rejection, confirming Stripe/Twilio operations now carry real body properties and every spec's eligible operations land in exactly one of tools/resources/resourceTemplates with no name collisions
* Live Petstore run: uploadImage (application/octet-stream) is now cleanly skipped with a warning instead of generating a broken tool
* pnpm run lint (prettier, eslint, tsc --noEmit) clean

### DIFF
--- a/docs/openapi.md
+++ b/docs/openapi.md
@@ -6,6 +6,7 @@
 
 - [Basic Usage](#basic-usage)
 - [Choosing which operations become tools](#choosing-which-operations-become-tools)
+- [GET → resources](#get--resources)
 - [Options](#options)
 - [Authentication](#authentication)
 - [Adding to an existing server](#adding-to-an-existing-server)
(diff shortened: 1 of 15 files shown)


punkpeye:
Pushed 41b2cea on top: a form-urlencoded body declared as a `$ref` (FastAPI `Body_*`, Box OAuth token/refresh/revoke) was skipped with an "unsupported content type" warning. Now resolved against `components.schemas` and flattened like an inline object; warning reworded. Merging once CI is green.

PR #351, opened 2026-09-02
feat(openapi): wire outputSchema/structuredContent, add query/body serialization styles, and a live resource-read test

LordNayan:
## Summary

Three additions to `fromOpenAPI()`, verified against the full real-spec benchmark suite (Petstore, Box, Twilio, PostHog, Stripe, Slack) — which is what surfaced two of the bugs below.

- **`outputSchema`/`structuredContent`**: a tool whose operation declares a `2xx` JSON response schema now returns typed `structuredContent` on success. The response is pre-validated against that exact schema first; a mismatch or non-object response falls back to today's plain text. Nothing that works today can start failing because of this.
- **Query/body serialization styles**: `deepObject`, `spaceDelimited`, `pipeDelimited` query styles, and bracket-notation for nested form-body objects/arrays (e.g. Stripe's `metadata[key]=value`, previously JSON-stringified into a single wrong value).
- **Live resource-read test**: `resources: true` is now exercised against real Petstore, closing the last gap where it was only ever tested offline or checked for tool/resource counts.

## Bugs found by testing at real scale

- **AJV crash on Box** — sample data under `example` coincidentally contained a field named `$id` (Box's own concept, not a JSON Schema keyword), reused across examples. AJV's `$id` discovery doesn't know `example` is documentation. Fixed by dropping `example`/`examples` when building schemas for AJV — they carry no validation meaning.
- **`tools/list` breakage from `outputSchema.type`** — the MCP SDK requires the literal string `"object"`. An unresolved `$ref` or a nullable object (`type: ["object", "null"]` after normalization) both failed that check and broke `tools/list` for every tool, not just the offending one. Fixed by resolving `$ref`s and normalizing the advertised type back to `"object"`.
- **320MB `tools/list` payload on Stripe** — the median operation's output schema pulled in 868 shared component schemas; uncapped, the full response would be ~320MB and the benchmark suite timed out past 150s. Capped at 50 referenced defs (~3MB, ~28s total).

## Known limitations (documented, not fixed)

- `explode: false` isn't respected — array query params are always exploded.
- A non-`deepObject` object-valued query/header/cookie parameter serializes as `"[object Object]"`.
- `resources: true` doesn't check response content-type, so a binary-returning `GET` could become a corrupted resource.

## Test plan

- [x] `pnpm test` — new unit tests incl. regressions for both `tools/list`-breaking cases and the defs cap
- [x] `pnpm test:openapi` — full suite passes against all 6 real specs (~28s, was timing out past 150s)
- [x] `pnpm run lint` clean

### DIFF
--- a/docs/openapi.md
+++ b/docs/openapi.md
@@ -7,6 +7,8 @@
 - [Basic Usage](#basic-usage)
 - [Choosing which operations become tools](#choosing-which-operations-become-tools)
 - [GET → resources](#get--resources)
+- [Structured output](#structured-output)
+- [Query and form-body serialization](#query-and-form-body-serialization)
 - [Options](#options)
 - [Authentication](#authentication)
 - [Adding to an existing server](#adding-to-an-existing-server)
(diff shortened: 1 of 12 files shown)


punkpeye (changes requested):
Solid work, and the benchmark-driven bug hunting is the right instinct. Four things before merge:

**1. `example`/`examples` are stripped from tool *input* schemas too — blocking.**
`STRIP_KEYS` is applied in `rewriteNode(…, "schema")`, which `buildFlatSchema`/`buildSharedDefs` also use. Verified on Petstore: `Order` loses `example: 10`, `example: "approved"`, etc. from the *input* schema. The AJV `$id` collision only affects the output path, so scope the strip to `buildOutputSchema`. Those examples are useful signal for a model filling arguments.

**2. `explode` is dead code.** `types.ts:133` adds `explode?: boolean`, never read anywhere in `src/`. The JSDoc above it also describes `style` (declared below it, undocumented). Drop the field and move the comment onto `style`.

**3. Misplaced JSDoc.** The `buildOutputSchema` doc block is attached to `const MAX_OUTPUT_SCHEMA_DEFS = 50`.

**4. `resolveToolResult` returns `Promise<string | unknown>`** — collapses to `unknown`; drop the union.

On `UNSTABLE_SPECS`: reasonable, but `box.json` has since drifted too (#356 re-pins both), so a posthog-only allowlist doesn't cover the case you're hitting. Fine to leave as-is here.

The rest — outputSchema wiring, the $ref/nullable normalization, the defs cap, serialization styles — looks right and is well tested.

punkpeye:
Pushed the changes to your branch rather than sending you round again — all four review points plus a merge of current `main`.

**The `example` strip is now opt-in and only `buildOutputSchema` opts in.** `rewriteNode` takes a `stripExamples` flag; input schemas keep their examples. Two call sites matter — the schema itself *and* the `$defs` it embeds, because `buildSharedDefs` runs once and is reused by the input path, so the strip has to happen at embed time.

I checked this was sufficient rather than assuming: reverting `STRIP_KEYS` reproduces your Box crash (`reference "01234500-…" resolves to more than one schema`), and scoping it to output-only keeps Box green. Worth noting the benchmark already AJV-compiles input schemas (line 124-127) and Box has always passed that on `main` — so the collision only ever came from the output path. Your diagnosis was right, just slightly wider than it needed to be.

Also: dropped the unused `explode` field and moved its JSDoc onto `style` (which it described), moved the `buildOutputSchema` doc block off `MAX_OUTPUT_SCHEMA_DEFS`, and `Promise<string | unknown>` -> `Promise<unknown>`.

Your `example`-strip test asserted the old input-path behaviour, so I retargeted it at `buildOutputSchema` and added three: `$defs` stripping, input schemas retaining examples, and `example`-as-a-field-name on the output path. 93 offline, 15/15 network, lint clean.

Nice work — the benchmark-driven bug hunting is what made the $id and tools/list issues findable at all. Merging once CI is green.

LordNayan:
This completes my 3 phase PR for Openapi to MCP. Thanks for constant review and collaboration @punkpeye

Issue #182, opened 2025-09-30
Bug: stateless: true mode only authenticates on initialize, not on tool calls


gazzadownunder:
# Bug: `stateless: true` mode only authenticates on initialize, not on tool calls

## Environment

- **FastMCP Version:** 1.0.0
- **Node Version:** 18.0.0+
- **Transport:** httpStream
- **Authentication:** JWT Bearer tokens (RFC 6750)

## Summary

FastMCP's `stateless: true` mode **only invokes the `authenticate()` callback on `initialize` requests, NOT on subsequent tool call requests** (`tools/list`, `tools/call`, etc.). This causes all tool calls to fail with "No valid session ID provided" even when valid JWT Bearer tokens are present in Authorization headers.

## Expected Behavior

When FastMCP is configured with `stateless: true`, the `authenticate()` callback should be invoked for **every incoming request** to validate authentication credentials and establish user context per-request (true stateless authentication).

```typescript
await server.start({
  transportType: 'httpStream',
  stateless: true,  // Should authenticate EVERY request
  authenticate: async (request) => {
    // Should be called for:
    // ✓ initialize request
    // ✗ tools/list request (currently NOT called)
    // ✗ tools/call request (currently NOT called)
    // ✗ resources/* requests (currently NOT called)
  }
});
```

## Actual Behavior

The `authenticate()` callback is **only invoked once during the `initialize` request**. All subsequent requests bypass authentication entirely and fail with session-related errors.

| Request Type | Expected | Actual |
|-------------|----------|--------|
| `initialize` | authenticate() called ✓ | authenticate() called ✓ |
| `tools/list` | authenticate() called ✓ | authenticate() NOT called ✗ |
| `tools/call` | authenticate() called ✓ | authenticate() NOT called ✗ |

### Setup

```typescript
import { FastMCP } from 'fastmcp';

class AuthServer {
  private server: FastMCP;

  constructor() {
    this.server = new FastMCP({ name: 'Auth Test', version: '1.0.0' });

    this.server.addTool({
      name: 'user-info',
      description: 'Get current user info',
      parameters: z.object({}),
      execute: async (args, context) => {
        console.log('[TOOL] user-info called');
        console.log('[TOOL] Context:', context);
        return JSON.stringify({ user: context.session });
      }
    });
  }

  private async authenticateRequest(request: any): Promise<any> {
    console.log('[AUTH] ========== Authentication Called ==========');
    console.log('[AUTH] Method:', request?.method);
[...]

Issue #180, opened 2025-09-28
Authentication failures should return 401 instead of 400


bokelley:
# FastMCP Authentication Issue

## Problem
When authentication is configured in FastMCP, requests without a valid session ID return:
- Status: 400 Bad Request
- Error: "No valid session ID provided"

## Expected Behavior
According to HTTP standards, missing authentication should return:
- Status: 401 Unauthorized
- Error: Authentication required

## Current Behavior
```bash
curl -v https://api.agentic.scope3.com/mcp
# Returns: HTTP/2 400
# Body: {"error":{"code":-32000,"message":"Bad Request: No valid session ID provided"}}
```

## Impact
- Violates HTTP semantics (400 is for malformed requests, 401 is for missing authentication)
- Confuses clients about whether the issue is with request format or authentication
- Makes it harder to handle authentication errors programmatically

## Suggested Fix
FastMCP should return 401 Unauthorized when:
1. No session ID is provided
2. Invalid session ID is provided
3. Session has expired

Only return 400 Bad Request for actual malformed requests.

PR #353, opened 2026-09-03
fix: invoke authenticate once per HTTP request on httpStream

### LINKED ISSUE #352: httpStream: authenticate is invoked twice per HTTP request (4.x; 3.13 invokes it once)
## Summary

On `httpStream`, FastMCP 4.x invokes the `authenticate` callback **twice for a single HTTP request**. 3.13 invokes it once. Both invocations receive the *same* `IncomingMessage` object, so this is one request being authenticated twice rather than two requests.

This matters when `authenticate` has any non-idempotent side effect. In our case it charges a fixed-window rate limiter, so every caller spent budget at twice the intended rate and the effective per-user limit was halved.

## Environment

- fastmcp 4.19.0 (also reproduced on `stateless: false`)
- mcp-proxy 6.7.11
- Node 22.15.0
- transport: `httpStream`

## Reproduction

```js
// repro.mjs
import { FastMCP } from 'fastmcp';
import { z } from 'zod';
import { readFileSync } from 'node:fs';

const version = JSON.parse(
  readFileSync(new URL('./node_modules/fastmcp/package.json', import.meta.url), 'utf8')
).version;
const PORT = 9030;

let calls = 0;
let firstRequest = null;
let sameObject = null;

const server = new FastMCP({
  name: 'repro',
  version: '1.0.0',
  authenticate: async (request) => {
    calls += 1;
    if (firstRequest === null) firstRequest = request;
    else sameObject = firstRequest === request;
    return { userId: 'repro' };
  },
});

server.addTool({
  name: 'ping',
  description: 'returns pong',
  parameters: z.object({}),
  execute: async () => 'pong',
});

const httpStream = { port: PORT };
if (version.startsWith('4.')) httpStream.stateless = true;
await server.start({ transportType: 'httpStream', httpStream });

// ONE request.
await fetch(`http://localhost:${PORT}/mcp`, {
  method: 'POST',
  headers: { 'Content-Type': 'application/json', Accept: 'application/json, text/event-stream' },
  body: JSON.stringify({
    jsonrpc: '2.0', id: 1, method: 'initialize',
    params: { protocolVersion: '2024-11-05', capabilities: {}, clientInfo: { name: 'repro', version: '1' } },
  }),
}).catch(() => {});

console.log(`fastmcp version:      ${version}`);
console.log(`HTTP requests sent:   1`);
console.log(`AUTHENTICATE CALLS:   ${calls}`);
if (calls > 1) console.log(`same IncomingMessage: ${sameObject}`);
process.exit(0);
```

## Result

The **same file**, changing only the installed version:

```
$ npm i fastmcp@4.19.0 zod && node repro.mjs
fastmcp version:      4.19.0
HTTP requests sent:   1
AUTHENTICATE CALLS:   2
same IncomingMessage: true

$ npm i fastmcp@3.13.0 && node repro.mjs
fastmcp version:      3.13.0
HTTP requests sent:   1
AUTHENTICATE CALLS:   1
```

Expected: 1. Actual on 4.x: 2.

## Cause

Both `httpStream` branches in `FastMCP#start` pass `authenticate` to `startHTTPServer` **and** call it again inside the `createServer` callback they hand to the same function:

```js
this.#httpStreamServer = await startHTTPServer({
  ...this.#authenticate ? { authenticate: this.#authenticate } : {},   // (1) mcp-proxy calls it
  createServer: async (request) => {
    let auth;
    if (this.#authenticate) {
      auth = this.#requireAuthenticated(await this.#authenticate(request));  // (2) called again
    }
    ...
  },
```

mcp-proxy invokes the `authenticate` option per request (`startHTTPServer.ts`, "Per-request authentication for all requests") and keeps the result as `authResult`; FastMCP then calls the callback a second time to obtain the same value for the session. The stateless and stateful branches both do this.

Presumably related to #182, which required `authenticate` to run on every request rather than only on `initialize` — the fix should keep that property.

## Suggested fix

Memoise per request rather than per session. Both calls receive the same `IncomingMessage`, so a `WeakMap` keyed on the request collapses them into one invocation while preserving per-request authentication (a new request is a new object). Storing the *promise* rather than the resolved value matters, because the second call can arrive before the first settles and should join it rather than start a competing attempt.

Happy to send a PR along those lines.

aaronik:
Fixes #352.

## The problem

On `httpStream`, `authenticate` is invoked **twice for one HTTP request**. `startHTTPServer` authenticates each request itself, and the `createServer` callback handed to it needs the same result to build the session — both were calling the consumer's hook. Stateless and stateful branches alike.

Harmless when `authenticate` is pure. Not harmless when it isn't: we charge a fixed-window rate limiter from it, so every caller spent budget at twice the intended rate and the effective per-user limit was halved.

3.13 invoked it once, so this is a regression relative to that line.

## The fix

Both invocations receive the *same* `IncomingMessage`, so a `WeakMap` keyed on the request collapses them into one call.

- **Per-request authentication is preserved.** A new request is a new object, so it authenticates again — the property #182 asked for still holds, and there is a test for it.
- **The in-flight promise is stored, not the resolved value.** The second caller can arrive before the first settles; it should join that attempt rather than start a competing one, which also means a rejection is shared rather than re-derived.
- Entries live exactly as long as the request.

## Tests

`src/FastMCP.authenticate-once.test.ts`, three cases:

- one request authenticates once, `stateless: true`
- one request authenticates once, `stateless: false`
- three requests authenticate three times — guards against over-caching, which would reintroduce #182

Reverting the source change fails all three (`expected [ Array(2) ] to have a length of 1`, and `expected 6 to be 3`), so they measure the fix rather than passing either way.

`prettier --check`, `tsc --noEmit` and `eslint` are clean on both changed files.

One note on the full suite: `src/FastMCP.test.ts` fails 2–3 tests with `EADDRINUSE` on ephemeral ports. That reproduces on a clean checkout of `main` with none of these changes, so it is pre-existing flakiness rather than something this PR introduces. The new test uses `getRandomPort()`, matching the convention in `FastMCP.routes.test.ts`, so it does not add to it.

### DIFF
--- a/src/FastMCP.authenticate-once.test.ts
+++ b/src/FastMCP.authenticate-once.test.ts
@@ -1,0 +1,115 @@
+import { getRandomPort } from "get-port-please";
+import { describe, expect, it } from "vitest";
+import { z } from "zod";
+
+import { FastMCP } from "./FastMCP.js";
+
+interface TestAuth {
+  [key: string]: unknown;
+  userId: string;
+}
+
+const initialize = (port: number) =>
+  fetch(`http://localhost:${port}/mcp`, {
+    body: JSON.stringify({
+      id: 1,
+      jsonrpc: "2.0",
+      method: "initialize",
+      params: {
+        capabilities: {},
+        clientInfo: { name: "test", version: "1.0.0" },
+        protocolVersion: "2024-11-05",
+      },
+    }),
+    headers: {
+      Accept: "application/json, text/event-stream",
+      "Content-Type": "application/json",
+    },
+    method: "POST",
+  }).catch(() => undefined);
+
+/**
+ * `startHTTPServer` authenticates each request, and the `createServer` callback
+ * handed to it needs the same result. Both used to call the consumer's hook, so
+ * one request authenticated twice — silent unless `authenticate` has a side
+ * effect, and destructive when it has one (a rate limiter charged there spends
+ * every caller's budget at twice the intended rate).
+ */
+describe("authenticate is invoked once per HTTP request", () => {
+  for (const stateless of [true, false]) {
+    it(`calls authenticate exactly once for one request (stateless: ${stateless})`, async () => {
+      const port = await getRandomPort();
+      const requests: unknown[] = [];
+
+      const server = new FastMCP<TestAuth>({
+        authenticate: async (request) => {
+          requests.push(request);
+
+          return { userId: "test-user" };
+        },
+        name: "authenticate-once",
+        version: "1.0.0",
+      });
+
+      server.addTool({
+        description: "returns pong",
+        execute: async () => "pong",
+        name: "ping",
+        parameters: z.object({}),
+      });
+
+      await server.start({
+        httpStream: { port, stateless },
+        transportType: "httpStream",
+      });
+
+      try {
+        await initialize(port);
+
+        expect(requests).toHaveLength(1);
+      } finally {
+        await server.stop();
+      }
+    });
+  }
+
+  it("authenticates each request separately rather than caching across them", async () => {
+    // The memo is keyed on the request object, so this must NOT collapse into a
+    // single authentication — that would reintroduce #182, where a stateless
+    // server authenticated `initialize` and then trusted everything after it.
+    const port = await getRandomPort();
+    let calls = 0;
+
+    const server = new FastMCP<TestAuth>({
+      authenticate: async () => {
+        calls += 1;
+
+        return { userId: "test-user" };
+      },
+      name: "authenticate-per-request",
+      version: "1.0.0",
+    });
+
+    server.addTool({
+      description: "returns pong",
+      execute: async () => "pong",
+      name: "ping",
+      parameters: z.object({}),
+    });
+
+    await server.start({
+      httpStream: { port, stateless: true },
+      transportType: "httpStream",
+    });
+
+    try {
+      await initialize(port);
+      await initialize(port);
+      await initialize(port);
+
+      expect(calls).toBe(3);
+    } finally {
+      await server.stop();
+    }
+  });
+});
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -5061,19 +5061,19 @@ test("authentication failure handling: returns 401 with WWW-Authenticate even wh
   const port = await getRandomPort();
   let callCount = 0;
 
-  // Regression test for a session-mode (non-stateless) request where
-  // `authenticate()` is invoked twice for the same request -- once by the
-  // transport layer and once internally by FastMCP when building the
-  // session. If the second call reports a failure whose message doesn't
-  // happen to contain a magic keyword (e.g. "Authentication", "Token",
-  // "Unauthorized"), the response must still be 401, not a generic 500.
+  // Regression test for a session-mode (non-stateless) request whose
+  // `authenticate()` reports a failure with a message that doesn't happen to
+  // contain a magic keyword (e.g. "Authentication", "Token", "Unauthorized").
+  // The response must still be 401 carrying the `WWW-Authenticate` challenge,
+  // not a generic 500 and not a bare 401.
+  //
+  // This used to return `{ authenticated: true }` on the first call and the
+  // failure on the second, because `authenticate()` ran twice for one request.
+  // It now runs once, so the failure is reported on that single invocation --
+  // which is the same path, reached without depending on two calls disagreeing.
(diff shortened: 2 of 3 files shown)


aaronik:
Marking this draft — CI caught something that makes the change incomplete as written, and it is worth stating precisely.

`FastMCP.test.ts > authentication failure handling: returns 401 with WWW-Authenticate even when the error message has no auth-related keywords` fails with this patch. It passes on a clean `main`. That test is not incidentally coupled to the double call — it depends on it deliberately, returning `{ authenticated: true }` on the first invocation and `{ authenticated: false, error: "Access denied" }` on the second, to reach the path where the *second* call's failure has to surface as 401 rather than 500 (#180).

Rewriting it to fail on a single invocation does not rescue it, and the reason is the actual problem. With one authentication, the failure is caught by `startHTTPServer` rather than by FastMCP's `createServer`. Both answer 401, but they do not answer identically:

- `startHTTPServer` sets `WWW-Authenticate` via `getWWWAuthenticateHeader(oauth, …)`, which returns `undefined` when no `oauth` is configured — so the header is simply absent.
- The `createServer` path produces the challenge regardless.

So with the second call removed, a server with no `oauth` config answers 401 with no `WWW-Authenticate` header. The duplicate invocation is load-bearing for that behaviour today. Measured, not inferred: the rewritten single-call test passes `expect(response.status).toBe(401)` and then fails on `expect(response.headers.get("WWW-Authenticate")).toContain("Bearer")` with the header `null`.

That leaves a design choice that is yours rather than mine:

1. `startHTTPServer` emits a default `Bearer` challenge on a 401 even without `oauth` configured, and FastMCP stops calling `authenticate` a second time.
2. FastMCP passes its own OAuth configuration down in more cases, so the proxy can build the challenge.
3. The double call stays, and the double-charging in #352 is addressed some other way — documenting that `authenticate` must be idempotent would at least make it a contract rather than a surprise.

Happy to implement whichever you prefer. The underlying report in #352 stands either way: one HTTP request currently authenticates twice, 3.13 authenticated once, and any non-idempotent work in the hook is silently doubled.

aaronik:
Resolved, and out of draft. `Test (24)` is green.

The gap was real and worth stating plainly: collapsing the duplicate call moved the `{ authenticated: false }` envelope's 401 from FastMCP to `startHTTPServer`, and those two are not equivalent — the transport can only build a challenge when the consumer configured `oauth`, so a server without one answered 401 with no `WWW-Authenticate` header.

Rather than choose between the three options above, the check simply moves to where it can always answer correctly. `#createSession`'s envelope check is now `#throwIfUnauthenticated`, called on the single authentication, so the 401 is raised by FastMCP and carries the challenge exactly as before. `#createSession` still calls it, leaving the stdio and in-memory paths untouched. No change to mcp-proxy required.

The #180 regression test is updated: it reported the failure on the *second* call, which only worked because the hook ran twice. It now reports it on the one invocation and additionally pins that the hook ran exactly once. All three of its original assertions are unchanged — 401, a Bearer challenge, and the verbatim `Access denied` message.

Reverting `FastMCP.ts` to `main` while keeping the tests fails all four (the three new ones, plus that #180 test), so they measure the change rather than passing either way.

On the remaining red check: `OpenAPI (network)` fails with `box.json does not match its pinned hash`. That reproduces on a clean checkout of `main` with none of these changes — Box's published spec has drifted since the hash was pinned, so it is unrelated to this PR and looks like a job for `npm run test:openapi:refresh` rather than something to fold in here. Happy to leave it or pick it up separately, whichever you prefer.

punkpeye:
Closing in favour of #355, which fixes the same bug with the same WeakMap-on-`IncomingMessage` mechanism but a narrower diff. Thanks for the clear writeup and repro — the diagnosis was right.

The deciding difference is that moving the envelope check into `#authenticateOncePerRequest` makes FastMCP's own 401 win, and `#createUnauthorizedResponse` hardcodes `id: null` where mcp-proxy echoes `body?.id ?? null`. Verified against main: main and #355 return `"id":1`, this branch returns `"id":null`. JSON-RPC 2.0 reserves `null` for an undeterminable id, so clients correlating by id break.

Two smaller things: `#throwIfUnauthenticated` ends up with two stacked JSDoc blocks, the first of which describes `#requireAuthenticated` (left undocumented); and the check move is a behaviour change beyond the reported bug.

You did surface something real though: a 401 without `WWW-Authenticate` violates RFC 7235, and that's true on main today for servers with no `oauth` config. Opening a follow-up — a PR that adds the challenge *and* preserves the request id would be very welcome.

PR #360, opened 2026-09-05
fix(http): settle failing custom route streams

pacocartones:
## Summary

- treat a custom route as handled once its response has started
- log and end the response if streaming fails after headers are sent
- avoid falling through to MCP transport routing and attempting a second response
- add a regression for a stream that errors after its first chunk

A Hono response body can reject asynchronously after the first chunk. At that point `respondWithCustomRoute()` cannot send an error response, but returning `false` tells the outer handler that no custom route matched. The request then falls through and can hang or attempt to write a second response.

## Validation

- regression test times out on the parent commit and passes with this change
- 23 route tests pass
- full lint, type checking and JSR publish dry-run pass

### DIFF
--- a/src/FastMCP.routes.test.ts
+++ b/src/FastMCP.routes.test.ts
@@ -1102,3 +1102,51 @@ test("custom route stream stops when a quiet client disconnects", async () => {
     await server.stop();
   }
 });
+
+test("custom route stream failure after headers are sent settles the response", async () => {
+  const port = await getRandomPort();
+  const server = new FastMCP({
+    name: "Test",
+    version: "1.0.0",
+  });
+  const unhandledRejections: unknown[] = [];
+  const onUnhandledRejection = (reason: unknown) => {
+    unhandledRejections.push(reason);
+  };
+  process.on("unhandledRejection", onUnhandledRejection);
+
+  const app = server.getApp();
+  app.get("/failing-stream", () => {
+    const encoder = new TextEncoder();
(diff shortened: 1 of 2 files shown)


punkpeye (changes requested):
Stops the fall-through correctly. One change needed: settle with `destroy()`, not `end()`.

Verified with `res.end()`:
- chunked body: client gets the truncated body with a clean terminator, so it looks complete
- `Content-Length` set: client keeps waiting for the missing bytes, so the hang moves client-side

Replace the `if (!res.writableEnded) { res.end(); }` block with:

```ts
// Drop the connection so the client sees the failure. end() writes nothing
// after destroy() but sets writableEnded, which mcp-proxy checks before
// running its own fallback response.
res.destroy();
res.end();
```

Test:

```ts
await expect(
  fetch(`http://localhost:${port}/failing-stream`).then((r) => r.text()),
).rejects.toThrow();
```

Checked locally: both cases reject in ms, 24/24 route tests pass, no unhandled rejections.

punkpeye:
Pushed the `destroy()` change myself to land this.

- `res.destroy(); res.end()` — client now sees a socket error instead of a truncated body that looks complete
- test asserts rejection, and covers the `Content-Length` case too (that one hung the client)

Thanks for the fall-through fix — that part was right.

punkpeye (approved):
destroy() fix pushed and green.

PR #364, opened 2026-09-09
fix(auth): persist DCR clients before caching

pacocartones:
## Summary

- persist dynamically registered OAuth clients before exposing them through the in-memory cache
- apply the same ordering to Client ID Metadata Document registrations
- cover failed durable writes in both registration paths

If storage rejects a new client record today, the client is already cached and remains usable for the lifetime of the process even though registration failed. Persisting first keeps the durable store and cache consistent: failed writes leave no ghost client, while successful registrations retain the existing behavior.

## Validation

- complete lint chain: Prettier, ESLint, TypeScript and JSR dry-run
- complete test suite: 61 files, 764 tests


punkpeye:

PR #365, opened 2026-09-09
fix(edge): roll back failed session initialization

pacocartones:
## Summary

- remove a session when its initialization request rejects
- guard rollback by initialization generation so a late failure cannot delete a newer session reusing the same ID
- cover retry after failure and the late-failure/ABA case

The edge transport currently installs a session before initialization completes. A rejected initialization therefore leaves a stale session that blocks retry. A simple unconditional delete would introduce a second race: an older rejection could delete a replacement session. Tracking the initialization generation makes cleanup precise.

## Validation

- complete lint chain: Prettier, ESLint, TypeScript and JSR dry-run
- complete test suite: 61 files, 764 tests


punkpeye:

PR #86, opened 2025-05-31
feat: embedded resources API (`server.embedded(uri)` method)

fauziralpiandi:
Implements the embedded resources API, providing **a clean and simple way to include resources in tool responses** without ~code~ code duplication:

```ts
// Example usage
server.addTool({
  name: "get_user_data",
  execute: async (args) => {
    return {
      content: [{
        type: "resource",
        resource: await server.embedded(`user://profile/${args.userId}`)
      }]
    };
  }
});
```

Closes #85

PR #366, opened 2026-09-09
fix(resources): normalize embedded() results like resources/read

ramin-010:
`server.embedded()` doesn't normalise `load()` results the way the `resources/read` handler does.

When a resource template's `load()` returns an array — which the type allow n and `resources/read` handles — `embedded()` tests `"text" in result` against the array itself, so neither `text` nor `blob` is ever set. A resource with neither field isn't valid `EmbeddedResource` content, so the whole `tools/call` response fails schema validation and the caller gets `MCP error -32602: Invalid input: expected string, received undefined`, pointing at the response shape rather than the resource that caused it.

Two smaller divergences alongside it: for both direct resources and templates, `embedded()` ignores the `mimeType` and `uri` set on the load result. `resources/read` uses `result.mimeType ?? resource.mimeType` and `result.uri ?? resource.uri`.

A `load()` returning no contents previously either threw a `TypeError` (direct resources) or produced an invalid embedded resource (templates); both now fail with the same explicit error.

This normalises array results and adds the `mimeType` and `uri` fallbacks in both branches. `embedded()` was added in #86 and array-returning `load()` in #123 two weeks later; the helper was never updated for it.

Five tests, all failing on main.

### DIFF
--- a/src/FastMCP.test.ts
+++ b/src/FastMCP.test.ts
@@ -26,6 +26,7 @@ import {
   imageContent,
   SessionError,
   type TextContent,
+  UnexpectedStateError,
   UserError,
 } from "./FastMCP.js";
 import { getTestPort } from "./getTestPort.js";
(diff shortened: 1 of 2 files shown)


ramin-010:
The OpenAPI (network) failure isn't from this change. `ensureBenchmarkSpecs` fails its
hash check because Twilio republished `twilio_api_v2010.json` upstream, so the job aborts
during setup. I get the same failure with the same hashes on a clean `main` locally.

This PR only touches `src/FastMCP.ts` and `src/FastMCP.test.ts`, and `Test (24)` is green.
Leaving the re-pin to you since it's a separate concern.

For what it's worth, `posthog.yaml` drifted too but is allow-listed as transiently
unstable and only warns; `twilio.json` has no such allowance and hard-fails.

punkpeye:

ramin-010:
Checks are green now. Thanks for re-pinning the specs.

PR #379, opened 2026-09-17
update @apidevtools/swagger-parser

mrjono1:
This will resolve CVE-2026-15195 in the sub dependency `@apidevtools/json-schema-ref-parser`


punkpeye:
Merging — thanks.

Heads-up on the rationale: CVE-2026-15195 does not resolve in the GitHub Advisory DB (no match for the ID, and the only advisory on `@apidevtools/json-schema-ref-parser` is CVE-2024-29651, which affects 11.0.0–11.1.0 — we were on 14.0.1, unaffected).

The upgrade is still worth it, via `js-yaml`. It pulls the transitive from 4.1.0 → 4.3.2, which clears three HIGH advisories:

- CVE-2026-84375 (`>=4.0.0, <4.3.2`)
- GHSA-5p4m-2wfm-xmqj (`>=4.0.0, <4.3.1`)
- CVE-2026-59869 (`>=4.0.0, <4.3.0`)

Merged on that basis. One follow-up on my side: the lockfile was regenerated with pnpm 9, which dropped the 22 `libc:` annotations pnpm 10 writes — restoring those separately.

PR #385, opened 2026-09-19
fix(openapi): keep the body when a JSON schema has an empty properties map

### LINKED ISSUE #380: fromOpenAPI drops JSON dictionary bodies with empty properties
`fromOpenAPI()` drops a JSON dictionary request body when its schema contains
`properties: {}`. The same schema without `properties` exposes a `body` argument
and forwards the dictionary correctly. An empty property map does not forbid
additional properties in OpenAPI.

Reproduced on main `548c5b696d3321fcb11278af083233c5734c947f`, Windows,
Node 24.21.0. Both required and optional request bodies are affected.

For example, `zod-to-json-schema@3.25.0` with `target: "openApi3"` emits the
working shape for `z.record(z.string())` and the failing shape for
`z.object({}).catchall(z.string())`. Both accept the same string dictionary.

### Reproduction

Run this with FastMCP, its MCP SDK dependency, and the optional `ajv` dependency
available. The HTTP implementation is a stub; it does not call an external API.

```js
import { Client } from "@modelcontextprotocol/sdk/client/index.js";
import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
import { fromOpenAPI } from "fastmcp/openapi";

for (const emptyProperties of [false, true]) {
  const requests = [];
    fetch: async (url, init) => {
      requests.push({ url, body: init.body });
      return new Response("{}");
    },
    spec: {
      openapi: "3.0.3",
      info: { title: "Labels", version: "1.0.0" },
      servers: [{ url: "https://api.example.test" }],
      paths: {
        "/labels": {
          post: {
            operationId: "setLabels",
            requestBody: {
              required: true,
              content: {
                "application/json": {
                  schema: {
                    type: "object",
                    additionalProperties: { type: "string" },
                    ...(emptyProperties ? { properties: {} } : {}),
                  },
                },
              },
            },
            responses: { 200: { description: "OK" } },
          },
        },
      },
    },
  });
  const client = new Client({ name: "repro", version: "1.0.0" });
  const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
  try {
    await Promise.all([
      server.connect(serverTransport),
      client.connect(clientTransport),
    ]);
    await client.listTools();
    await client.callTool({
      name: "setLabels",
      arguments: { body: { team: "infra" } },
    });
    console.log({ emptyProperties, requests });
  } catch (error) {
    console.log({ emptyProperties, error: error.message, requests });
  } finally {
    await server.sessions[0]?.close();
    await client.close();
    await server.stop();
  }
}
```

Without `properties`, the request body is `{"team":"infra"}`. With
`properties: {}`, `callTool()` rejects with MCP `-32602`, reporting
`Tool 'setLabels' parameter validation failed: root: must NOT have additional properties`.
No HTTP request is made in that case.

Expected: both schemas retain a whole-body argument, validate dictionary values
against the declared schema, and send the same JSON payload.

### Cause and boundary

`extractBodyProperties()` in `src/openapi/schemas.ts` treats the existence of
`schema.properties` as enough to flatten an object. An empty map contributes no
arguments, losing the entire request body. The existing whole-body path already
handles the equivalent dictionary without `properties`.

This is separate from the documented policy that FastMCP advertises strict
input schemas in `tools/list`. The reproduction concerns runtime argument
mapping and forwarding; changing that global schema policy is outside this fix.

PerryLink:
## What

`fromOpenAPI()` no longer drops a JSON request body whose schema carries an empty `properties` map.

## Why

A dictionary-like body — `additionalProperties` with no fixed keys — can be emitted as `properties: {}`. For example `zod-to-json-schema` with `target: "openApi3"` produces that shape for `z.object({}).catchall(z.string())`, while `z.record(z.string())` produces a schema with no `properties` key at all.

In `buildFlatSchema`, the object branch is guarded by `schema.type === "object" && schemaProperties`. An empty object is truthy, so the branch is taken, `Object.entries` yields nothing to flatten, and the function returns early — before the fallback that sets `wholeBodyKey: "body"`. The tool ends up with no body argument at all, which is why the same schema without the `properties` key works while this one silently drops the payload.

An empty property map is not the same as an absent one, and it does not forbid additional properties in OpenAPI.

Fixes #380.

## Change

`src/openapi/schemas.ts` — flatten only when there is at least one property. An object schema with an empty `properties` map now reaches the existing whole-body branch and is exposed as a single `body` argument carrying the original schema, so `additionalProperties` is preserved. No other branch changes; schemas with real properties still flatten exactly as before.

`src/openapi/schemas.test.ts` — two regression tests: an empty `properties` map exposes `body` with a populated `parameterMap`, and a required one is listed in `required`.

## Validation

Reproduced before the fix through `buildFlatSchema` directly, on `main` at `548c5b69`:

| body schema | `properties` before | `parameterMap` before |
| --- | --- | --- |
| `{ type: "object", properties: {} }` | `[]` | `{}` — body dropped |
| `{ type: "object" }` (no `properties` key) | `["body"]` | `{"body":{"in":"body","name":"body"}}` |

After the change the first row behaves like the second.

```bash
npx vitest run src/openapi/schemas.test.ts   # 42 passed (40 before this change)
npx vitest run                               # 806 passed | 1 skipped | 1 failed (808)
```

The single failure is `FastMCP.test.ts > adds tools with outputSchema`. It fails identically on an unmodified checkout of `548c5b69`, so it is pre-existing and unrelated to this change.

Not verified: an end-to-end run through `fromOpenAPI()` against a `zod-to-json-schema`-generated document, and an actual request carrying the dictionary payload. The reproduction was reduced to the branch that drops it.


punkpeye:
Thanks — solid diagnosis, and it matches #380 exactly. Went with #381 for the fix.

Reason: your guard drops the \`schemaProperties\` truthiness check entirely, so a **form-encoded** body with \`properties: {}\` now falls past the object branch into the \`bodyEncoding === "form"\` early-return at `src/openapi/schemas.ts:434`. That sets \`unsupportedBodyContentType\`, and `fromOpenAPI.ts:93` skips the operation — the whole route disappears from the server. Verified locally:

| | before | after your patch |
|---|---|---|
| form body, \`properties: {}\` | route registered, no body args | `unsupportedBodyContentType` → **route skipped** |

#381 keeps `bodyEncoding === "form"` (and `additionalProperties: false`) on the flatten path, so only true JSON dictionaries reroute to whole-body.

No issue with your analysis of the JSON case — it was correct.

PerryLink:
You're right, and thank you for working through the mechanism rather than just closing it.

I re-checked this before replying, against the API-exact blob at each revision — each file's `git hash-object` matches the blob SHA the API reports for it, so the comparison is byte-exact rather than a re-typed copy:

```
                       form body, properties: {}          json body, properties: {}
base                   bodyEncoding:"form", props:[]       props:[]         <- the bug
#385                   unsupportedBodyContentType:...      props:["body"]
#381 (== main today)   bodyEncoding:"form", props:[]       props:["body"]   <- matches base on the form axis
```

So the form-encoded route does disappear under my patch, #381 does not have that failure mode, and #380 is genuinely fixed by it. Both of your line references are exact at the base revision: `schemas.ts:434` is the `if (bodyEncoding === "form") {` early return, and `fromOpenAPI.ts:93` is the guard that `continue`s before `server.addTool`.

One correction to your wording, offered in your favour rather than against it. The guard was not dropped, it was narrowed — from `schemaProperties` to `schemaProperties !== undefined && Object.keys(schemaProperties).length > 0`. That distinction matters, because `!== undefined` is *looser* than truthiness for `null`, and it introduces a second failure that neither of us mentioned:

```
properties: null     base -> body arg     #385 -> TypeError: Cannot convert undefined or null to object     #381 -> body arg
```

`properties: null` is not valid OpenAPI, so that is robustness rather than a real regression. But it is a further sign that replacing the predicate was the wrong move: your version kept the original guard and added one term, and that term is precisely the one that was missing.

For the record on my side, my tests covered only the JSON case, which is exactly why they could not have caught this. Thank you for catching it.

